PR #497, opened 2022-11-04
Fix RTA call graph generation

JonasKlauke:
Fixed the Testcase to represent a correctly generated RTA call graph. Fixed the RTA call graph generation to correctly generate the call graph.


sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -133,20 +180,122 @@ final void processWorkList(
...
+                if (rightOp instanceof JNewExpr) {
+                  Type type = rightOp.getType();
+                  if (type instanceof ClassType) {
+                    classType = (ClassType) type;
+                  }
+                } else if (rightOp instanceof JNewArrayExpr) {
+                  classType = findClassTypeInType(((JNewArrayExpr) rightOp).getBaseType());
+                } else if (rightOp instanceof JNewMultiArrayExpr) {
+                  classType = findClassTypeInType(((JNewMultiArrayExpr) rightOp).getBaseType());
+                }
```
swissiety: shall we use a ExprVisitor to make the ifelse O(1) ?

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -133,20 +180,122 @@ final void processWorkList(
...
+   * @param classType the type of the class that might contain the clinit method
+   * @return an optional of a SootMethod. The SootMethod is contained if the class described by the
+   *     class type contains an implementation of clinit, otherwise it is empty
+   */
+  private Optional<? extends SootMethod> findClinitMethodOfClassType(
+      @Nonnull View<? extends SootClass<?>> view, @Nonnull ClassType classType) {
+    MethodSignature clinitSig =
+        new MethodSignature(
+            classType,
+            new MethodSubSignature("<clinit>", Collections.emptyList(), VoidType.getInstance()));
```
swissiety: can we refactor that into ClassType.createClinitSignature() or into the JavaIdentifierfactory

JonasKlauke: mmh I will create a getStaticInitializer() in ClassType.

JonasKlauke: I have removed the method and replaced it in its usages with classtype.getStaticInitializer

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
JonasKlauke: done. InstantiateClassValueVisitor implements a ValueVisitor

sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmUtil.java
```diff
@@ -154,60 +154,96 @@ public static Type toJimpleSignature(@Nonnull String desc) {
...
+  private static int countArrayDim(@Nonnull String desc) {
+    int nrDims = desc.lastIndexOf('[') + 1;
+    for (int index = 0; index < nrDims; index++) {
+      if (desc.charAt(index) != '[') {
         throw new AssertionError("Invalid array descriptor: " + desc);
       }
-      desc = desc.substring(idx + 1);
     }
-    Type baseType;
+    return nrDims;
```
swissiety, line 197: is the format check necessary? (i know it was there before)

JonasKlauke: Before the change, the check was only the last [ and the first [
I think the question is if we want to be robust.
This method is normally called with descriptions extracted from ASM, so there shouldnt be an issue with the format.
But this is also a public method so you could feed it with selfcreated descriptions.

sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmUtil.java
```diff
@@ -154,60 +154,96 @@ public static Type toJimpleSignature(@Nonnull String desc) {
...
-      if (desc.charAt(0) != '[') {
+      desc = desc.substring(nrDims);
+    }
+
+    Type baseType = toPrimitiveOrVoidType(desc).orElse(null);
+    if (baseType == null && desc.charAt(0) != 'L') {
+      throw new AssertionError("Unknown descriptor: " + desc);
+    }
+
+    if (baseType == null && desc.charAt(0) == 'L') {
```
swissiety: please merge the same checks on: baseType==null

swissiety: so the second half becomes the else part as well

sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmUtil.java
```diff
@@ -154,60 +154,96 @@ public static Type toJimpleSignature(@Nonnull String desc) {
...
+      if (desc.charAt(desc.length() - 1) != ';') {
+        throw new AssertionError("Invalid reference descriptor: " + desc);
+      }
+      String name = desc.substring(1, desc.length() - 1);
+      name = toQualifiedName(name);
+      baseType = JavaIdentifierFactory.getInstance().getType(toQualifiedName(name));
+    }
+    if (!(baseType instanceof JavaClassType) && desc.length() > 1) {
```
swissiety: the errormsg does not match instanceof type?

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -191,7 +299,23 @@ final <T extends Method> T findMethodInHierarchy(
...
+   * @param workList the current worklist that might be extended
+   * @param cg the current cg that might be extended
+   */
+  public void preProcessingMethod(
+      View<? extends SootClass<?>> view,
+      MethodSignature sourceMethod,
+      @Nonnull Deque<MethodSignature> workList,
+      @Nonnull MutableCallGraph cg) {
+    // This is only implemented if it is needed in the call graph algorithm
+  }
```
kadirayk: why not make this and postProcessing abstract?

kadirayk (changes requested):

sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmUtil.java
JonasKlauke: Nope there are 4 different case :) 
But yeah I can put them into a if (baseType==null)....

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
JonasKlauke: then you have to create this dummy method for all other algorithm which are not RTA
But yeah we can do that, so that you are aware of this steps in the algorithm

kadirayk: I think it's better to make it explicit.

sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmUtil.java
JonasKlauke: no it matches. If basetype is not of instance JavaClassType, it is a primitive datatype. 
But all primitive datatyped are described with one Letter.
This Check is to prevent that "Cool" is not translated to Char primitive because the first Letter is C

JonasKlauke: but yeah matching it directly to Primitive Datatype might be more robust. But I think then Void is excluded because it is not a Primitive Datatype

kadirayk (approved):

PR #521, opened 2022-12-14
improvements Localgenerator

swissiety:


sootup.core/src/main/java/sootup/core/jimple/basic/LocalGenerator.java
```diff
@@ -64,138 +61,152 @@ public Local generateThisLocal(@Nonnull Type type) {
...
+      defaultCaseType();
+    }
 
-  private void appendNextRefLikeTypeName(StringBuilder name) {
-    name.append("r").append(tempRefLikeType++);
-  }
+    @Override
+    public void caseVoidType() {
+      // TODO: how does a local with a voidtype make sense..? but obviously there was code/ a letter
+      // assigned for it in old soot..
```
stschott: As far as I can recall, I've never seen a local with "voidtype".

stschott (approved):

sootup.core/src/main/java/sootup/core/jimple/basic/LocalGenerator.java
```diff
@@ -64,138 +61,152 @@ public Local generateThisLocal(@Nonnull Type type) {
...
+      type.accept(ns);
+      localCandidate = Jimple.newLocal(name.toString(), type);
+    } while (locals.contains(localCandidate));
+
+    locals.add(localCandidate);
+    return localCandidate;
   }
 
   /** generates a new {@link Local} given the type for local. */
   public Local generateLocal(@Nonnull Type type) {
```
kadirayk, line 80: - Wouldn't it be better to use a common method for generateField and generateLocal?
- I think do while in generateField looks cleaner than what's done here
- Also what about naming generateField to generatFieldLocal? 😄

swissiety: 1+2) I thought we could trade some decline of simplicity/readability for removing unnecessary runtime overhead as its not that complicated here
3) nice, I agree

kadirayk (approved):

PR #530, opened 2022-12-30
Fix indentation of Jimple code in doc

thebesttv:
The Jimple code in section [JReturnStmt & JReturnVoidStmt](https://soot-oss.github.io/SootUp/jimple/#jreturnstmt-jreturnvoidstmt) seems to have two spaces missing, which causes it to be rendered wrong:


ericbodden:
Thanks a lot!

ericbodden (approved):

PR #534, opened 2023-01-01
Fix `java.io.FileNotFoundException`

niorgai:
1. Rename the method.
2. Create the property directory if it does not exist, avoiding `java.io.FileNotFoundException`


swissiety (approved):
:+1:

swissiety:
maybe - we could move the properties file to a better path i.e. a directory that is not maven configuration dependent and would not need the proposed fix?

niorgai:
> maybe - we could move the properties file to a better path i.e. a directory that is not maven configuration dependent and would not need the proposed fix?


Agree, but I am not familiar with maven, I just encountered this problem when running the code.
If there is a better solution, it would be better

JRoper18:
Hey, I'm also running into this issue. Can this be merged?

kadirayk (approved):

FZhg:
thanks a lot!

PR #465, opened 2022-04-14
Type_Assigner

wangzun66:
Finished:
- [x] Typing
- [x] Intermediate Types 
- [x] PrimitiveHierachy + Test
- [x] Check logic of ViewTypeHierachy, and move it to soot.core 
- [x] Hierachy - BytecodeHierachy + Test
- [x] AugEvalFunction + Test
- [x] CastCounter
- [x] PromotionCounter
- [x] TypeResolver + test
- [x] TypeAssigner

closes #238

PR #540, opened 2023-01-18
fix WALA

JonasKlauke:


swissiety:
fixed the SwitchTests which now have the previously missing Locals from WALA  in #465

kadirayk (approved):

PR #546, opened 2023-02-08
check if parentFile exists

stschott:


sootup.java.sourcecode/src/main/java/sootup/java/sourcecode/frontend/WalaJavaClassProvider.java
```diff
@@ -183,7 +183,10 @@ private void createWalaProperties() {
     if (!walaPropertiesFile.exists()) {
       PrintWriter pw;
       try {
-        walaPropertiesFile.getParentFile().mkdirs();
+        File parentFile = walaPropertiesFile.getParentFile();
+        if (parentFile != null) {
+          parentFile.mkdirs();
+        }
```
swissiety, line 189: oh both PRs got merged.. with the new path i.e. just the file the parent directory (which is on my machine the sourcecodefrontend dir) should always exist so we can remove that check

PR #547, opened 2023-02-14
fix LocalSplitter testcase

### LINKED ISSUE #488: BodyInterceptor: fix LocalSplitter

swissiety:
closes #488


sootup.core/src/main/java/sootup/core/graph/MutableBlockStmtGraph.java
```diff
@@ -854,6 +854,10 @@ public void insertBefore(
     }
   }
 
+  /**
+   * Adds stmtB to StmtA and if B is not already in the StmtGraph assumes stmtB has the same
```
stschott: stmtA should be lower case for consistency

stschott (approved):

PR #549, opened 2023-02-17
[fix] Fix LocalPacker crashing when handling FieldRefs

### LINKED ISSUE #548: LocalPacker crashes with programs that contain FieldRefs
I've got a fix, I think. Just trying to figure out how to make a PR.

JRoper18:
The LocalPacker BodyInterceptor crashes with a NullPointerException when given any body with FieldRefs on the LHS of the assignment in it. This is mainly because it attempts to replace the FieldRef's local base value with BodyUtils.withNewUse(), which is not capable of doing usage replacements on the LHS of an assignment statement. This PR:

1. Fixes the crash
2. Adds a test that the issue is fixed
3. Extends the local packer to work for these cases. 

Closes #548

### DIFF
--- a/sootup.core/src/main/java/sootup/core/jimple/visitor/ReplaceUseStmtVisitor.java
+++ b/sootup.core/src/main/java/sootup/core/jimple/visitor/ReplaceUseStmtVisitor.java
@@ -69,21 +69,21 @@ public class ReplaceUseStmtVisitor extends AbstractStmtVisitor<Stmt> {
     }
   }
 
-  private void caseAssignRVStmtHelper(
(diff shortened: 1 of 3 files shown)


swissiety:
Thx, for your contribution! The problem with the suggested solution is the LHS makes it the def and not the use so we need another approach to replace it in a higher layer replacing the def (as this is in the Replace**Use**SmtVisitor). in #465 is a implementation to replace Locals on both sides i.e. for def and use.

JRoper18:
But doesn't "GetUses" return values in the LHS as well? In "AbstractDefinitionStmt"

JRoper18:
Oh, I see -- the PR linked removes that.

PR #582, opened 2023-04-06
move format check before compilation

stschott:
Move the format check before compilation, so the pipeline does not keep running and then fails because of formatting.
Rather fail directly before running the pipeline at all.


swissiety:
:+1: saves computing minutes
:-1: (unit) tests needs to be triggered again with another commit/push while hopefully just the format is wrong otherwise it needs to run again anyways.. just unlucky when your not doing busy waiting

the way the change is done is clean but is more time expensive as another checkout/vm is necessary - you could run the command in the compilation job as well

PR #611, opened 2023-05-30
Store Visible Annotations

### LINKED ISSUE #523: Store also visible annotations of methods
I called JavaSootMethod.getAnnotations and return no annotation, but it actually has many annotations.
image
In ByteCode, Here are many annotations.
image

palaniappan1:


JonasKlauke:
Hi, to get rid of the failing formating, go to the the Maven Tab in IntelliJ, SootUP -> Plugins -> fmt  and double click on fmt:format. That will fix all format issues and you can create a fmt commit

JonasKlauke (approved):

kadirayk (approved):

PR #618, opened 2023-06-06
Dot Exporter and Write to a file

### LINKED ISSUE #612: Dotexporter for call graphs
The call graph object should be able to export itself into the Dot format.
It should provide a method to export the content of a Dot file as a String
and it should be able to create a file with a given path and file options.

palaniappan1:


sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -23,10 +23,11 @@
...
 import com.google.common.base.Preconditions;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.*;
```
JonasKlauke: Change import java.util.*; back to 
import java.util.Comparator;
import java.util.HashMap;
import java.util.Map;
import java.util.Set;

sootup.callgraph/src/main/java/sootup/callgraph/CallGraph.java
```diff
@@ -86,6 +86,9 @@ boolean containsCall(
    */
   int callCount();
 
+  /** This method converts the call graph object into dot format and write it to a string file. */
+  void convertToDotFormatAndFileWrite(String filePath);
```
JonasKlauke: exportAsDot() is a better name

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -129,6 +130,84 @@ public int callCount() {
...
+  @Override
+  public void convertToDotFormatAndFileWrite(String filePath) {
+    StringBuilder dotFormatBuilder = new StringBuilder();
+
+    for (Edge edge : graph.edgeSet()) {
+      Vertex sourceVertex = graph.getEdgeSource(edge);
+      Vertex targetVertex = graph.getEdgeTarget(edge);
+      dotFormatBuilder
+          .append("\t")
+          .append(sourceVertex.methodSignature)
```
JonasKlauke: put it in "

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -129,6 +130,84 @@ public int callCount() {
...
+          .append(" -> ")
+          .append(targetVertex.methodSignature)
```
JonasKlauke: put it in ""

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -129,6 +130,84 @@ public int callCount() {
...
+          .append(";\n");
+    }
+    // Sort the callGraph
+    String sortedCallGraph = sortCallGraph(dotFormatBuilder.toString());
```
JonasKlauke: dont sort the string. Sort the edges in the call graph it is way easier

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -129,6 +130,84 @@ public int callCount() {
...
+        new Comparator<String>() {
+          @Override
+          public int compare(String line1, String line2) {
+            String className1 = extractClassName(line1);
+            String className2 = extractClassName(line2);
+
+            if (className1.equals(className2)) {
+              String methodName1 = extractMethodName(line1);
+              String methodName2 = extractMethodName(line2);
+              return methodName1.compareTo(methodName2);
```
JonasKlauke: missing comparision of parameters

JonasKlauke (changes requested):

sootup.callgraph/src/main/java/sootup/callgraph/CallGraph.java
swissiety: is there a method which returns it as String (as well)?

palaniappan1: Now we changed the method to just return a dot formatted string and the file write feature is removed.

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -129,6 +133,93 @@ public int callCount() {
...
+                  .append("\"" + sourceVertex.methodSignature + "\"")
+                  .append(" -> ")
+                  .append("\"" + targetVertex.methodSignature + "\"")
+                  .append(";\n");
+            });
+    // Sort the callGraph
+    String sortedCallGraph = sortCallGraph(dotFormatBuilder.toString());
+
+    // Using only one string builder, so deleting everything and adding the sorted callgraph
+    dotFormatBuilder.delete(0, dotFormatBuilder.capacity());
```
JonasKlauke: Don't do this. 
Instead just return the string concatination 
"strict digraph ObjectGraph {\n" + sortedCallgraph + "}"

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -129,6 +133,93 @@ public int callCount() {
...
+  }
+
+  /**
+   * Unlike toString method in the same file, the comparison is done manually by comparing two
+   * strings, because the details of className, MethodSignature is not present in the data at hand.
+   *
+   * @param dotOutput
+   * @return
+   */
+  public String sortCallGraph(String dotOutput) {
```
JonasKlauke: why do you still use this method?

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -129,6 +133,93 @@ public int callCount() {
...
   }
 
+  @Override
+  public String exportAsDot() {
+    StringBuilder dotFormatBuilder = new StringBuilder();
+
+    graph.edgeSet().stream()
+        .sorted(
+            Comparator.comparing(
+                    (Edge edge) -> graph.getEdgeSource(edge).toString(), Comparator.naturalOrder())
```
JonasKlauke: dont sort the entire signature.
otherwise the order will be class name , return type, method name and then parameter.
but we dont want that the return type highly effects the order. So you have to compare the parts of the method signatures

JonasKlauke (changes requested):

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
JonasKlauke: deleting the string builder seems unnecessary complex, and it might waste execution time

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
JonasKlauke: your edges are already kind of sorted. why sort them again?

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
JonasKlauke: wasting execution time to extract strings if we have already ways to get them before we call tostring on the method signature

JonasKlauke (approved):
Nice work 👍

PR #638, opened 2023-06-29
Minor Change in exportAsDot

palaniappan1:


sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -140,7 +140,7 @@ public String exportAsDot() {
             Comparator.comparing(
                     (Edge edge) -> {
                       Vertex edgeSource = graph.getEdgeSource(edge);
-                      return edgeSource.methodSignature.getDeclClassType().getClassName();
+                      return edgeSource.methodSignature.getDeclClassType().toString();
```
kadirayk: getFullyQualifiedName would be easier to understand, so we don't need to check what toString returns.

kadirayk (approved):

PR #640, opened 2023-06-30
Icfg dot exporter

### LINKED ISSUE #564: DotExporter for ICFG
Would it be difficult to get a printer like [DotExporter](https://github.com/soot-oss/SootUp/blob/develop/sootup.core/src/main/java/sootup/core/util/DotExporter.java) for this? Essentially, it would build the CFG for all methods and then add edges between callees and callers?

_Originally posted by @will-leeson in https://github.com/soot-oss/SootUp/issues/550#issuecomment-1456839966_

palaniappan1:
The test does not seem to make sense, as I am comparing two strings that I know will be the same. I tried the resulting string in the https://dreampuf.github.io/GraphvizOnline/ website to visualize the ICFG. I also attach the resulting png image of the ICFG.
graphviz


sootup.analysis/src/main/java/sootup/analysis/interprocedural/icfg/JimpleBasedInterproceduralCFG.java
```diff
@@ -148,6 +150,21 @@ public JimpleBasedInterproceduralCFG(
     initializeStmtToOwner();
   }
 
+  public String buildICFGGraph() {
+    ArrayList<StmtGraph> stmtGraphSet = new ArrayList<>();
+    // To Sort the methodSignature set with entrypoint as the first element.
+    Set<MethodSignature> methodSignatures = cg.getMethodSignatures();
+    methodSignatures.remove(mainMethodSignature);
+    Set<MethodSignature> modifiedMethodSignatures = new LinkedHashSet<>();
```
kadirayk: We don't need to get all the method signatures upfront. you can start from the main method and traverse the call graph using `cg.getCallsFrom()`. You can then add stmtGraphs to the dot graph as you traverse and add corresponding edges that represent the calls.

sootup.analysis/src/test/java/sootup/analysis/interprocedural/ifds/ICFGCallGraphTest.java
```diff
@@ -0,0 +1,112 @@
...
+    SootClass<?> sc = view.getClass(mainClassSignature).get();
+    entryMethod =
+        sc.getMethods().stream().filter(e -> e.getName().equals("entryPoint")).findFirst().get();
+
+    entryMethodSignature = entryMethod.getSignature();
+
+    JimpleBasedInterproceduralCFG icfg =
+        new JimpleBasedInterproceduralCFG(view, entryMethodSignature, false, false);
+    String expectedCallGraph =
+        "digraph G {\n"
```
JonasKlauke: Maybe instead of comparing the full result. you could check if all methods blocks are contained in the block and that they are connected by the calls.
Maybe you remove found parts from the string till the string is empty or you cant find a edge or code block in it.
this way it might be robust for small changes in order or something

sootup.analysis/src/test/resources/callgraph/ICFGExample.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public String thirdMethod(String var1){
+        return var1;
+    }
+
+    public void entryPoint() {
+        String var1 = "SECRET";
+        String var2 = this.id(var1);
+        this.sink(var2);
```
JonasKlauke: test also a case where you can have multiple call targets for one method call. Like you call method m which can go to Superclass m and Subclass m

JonasKlauke (changes requested):

sootup.analysis/src/test/resources/callgraph/ICFGExample.java
JonasKlauke: but do this in another testcase.

sootup.analysis/src/main/java/sootup/analysis/interprocedural/icfg/JimpleBasedInterproceduralCFG.java
palaniappan1: CallGraph is not accessible in ICFGDotExporter.java, I need to add a dependency for accessing callgraph. So this is the plausible way of doing it I reckon. I can't use cg.callsFrom() because I need the hashcode of the invoke statement.

kadirayk: What do you mean it's not accessible? You are calling `cg.getMethodSignatures()` ?

kadirayk (approved):

PR #648, opened 2023-07-15
Improve/unify identifier factory references

swissiety:
`Project` stored the `IdentifierFactory` reference via Language field *and* as a dedicated field.
Now only the Language is used to store it.

### DIFF
--- a/sootup.core/src/main/java/sootup/core/jimple/common/expr/JNegExpr.java
+++ b/sootup.core/src/main/java/sootup/core/jimple/common/expr/JNegExpr.java
@@ -69,17 +69,18 @@ public final class JNegExpr extends AbstractUnopExpr implements Copyable {
   public Type getType() {
     Value op = getOp();
 
-    if (op.getType().equals(PrimitiveType.getInt())
-        || op.getType().equals(PrimitiveType.getByte())
-        || op.getType().equals(PrimitiveType.getShort())
-        || op.getType().equals(PrimitiveType.getBoolean())
-        || op.getType().equals(PrimitiveType.getChar())) {
+    // TODO ms: create a TypeVisitor for this -> O(1)
+    if (op.getType() == PrimitiveType.getInt()
+        || op.getType() == PrimitiveType.getByte()
+        || op.getType() == PrimitiveType.getShort()
+        || op.getType() == PrimitiveType.getBoolean()
+        || op.getType() == PrimitiveType.getChar()) {
       return PrimitiveType.getInt();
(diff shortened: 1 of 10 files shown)


sootup.core/src/main/java/sootup/core/jimple/common/expr/JNegExpr.java
```diff
@@ -69,17 +69,18 @@ public void toString(@Nonnull StmtPrinter up) {
   public Type getType() {
     Value op = getOp();
 
-    if (op.getType().equals(PrimitiveType.getInt())
-        || op.getType().equals(PrimitiveType.getByte())
-        || op.getType().equals(PrimitiveType.getShort())
-        || op.getType().equals(PrimitiveType.getBoolean())
-        || op.getType().equals(PrimitiveType.getChar())) {
+    // TODO ms: create a TypeVisitor for this -> O(1)
+    if (op.getType() == PrimitiveType.getInt()
```
JonasKlauke, line 73: Does == work here? 
because it is the same object?

sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/EmptySwitchEliminator.java
```diff
@@ -42,15 +42,13 @@ public class EmptySwitchEliminator implements BodyInterceptor {
   @Override
   public void interceptBody(@Nonnull Body.BodyBuilder builder, @Nonnull View<?> view) {
     // Iterate all stmts in the body
-
-    for (Stmt stmt : builder.getStmtGraph()) {
+    for (Stmt stmt : new ArrayList<>(builder.getStmtGraph().getNodes())) {
```
JonasKlauke, line 45: why?

swissiety: a) order is not relevant (faster node access) b) changing a collection while iterating over it -> Concurrenmodificationexception

sootup.core/src/main/java/sootup/core/jimple/common/expr/JNegExpr.java
swissiety: exactly - for thatreason we implemented it this way - and to save memory and as we have private ctors its no other way to get to true - except maybe via nasty reflection stuff making the ctor accessible etc

JonasKlauke (approved):

PR #666, opened 2023-08-11
Fix/invoke dynamics

### LINKED ISSUE #619: Invokedynamic failures
Hi there, 

I’ve been investigating some failures parsing invokedynamic instructions. I first saw a similar issue raised in #574.

Just noting that I originally encountered the problem in v1.0.0 but it persists in v1.1.0. 

I ran into some issues parsing record classes. I was getting various stack underrun errors when attempting to get the bodies of these methods. For example, the record class generates an equals(), hashCode() and string() method by default - all of which trigger an indy. 

```
Public record RecordTest(String a, String b) {
	public RecordTest(String a, String b) {
		this.a = a; 
		this.b = b; 
	}
}

```

When I attempt to get the method body for equals(), I get the following stack trace: 
```
Exception in thread "main" java.lang.RuntimeException: Failed to convert <jlink.RecordTest: boolean equals(java.lang.Object)>
	at sootup.java.bytecode.frontend.AsmMethodSource.resolveBody(AsmMethodSource.java:201)
	at sootup.core.model.SootMethod.lazyBodyInitializer(SootMethod.java:95)
	at com.google.common.base.Suppliers$NonSerializableMemoizingSupplier.get(Suppliers.java:166)
	at sootup.core.model.SootMethod.getBody(SootMethod.java:134)
	at org.ifds.IFDSSetUp.setupSoot(IFDSSetUp.java:68)
	at org.ifds.IFDSSetUp.executeStaticAnalysis(IFDSSetUp.java:30)
	at org.ifds.Main.main(Main.java:15)
Caused by: java.lang.RuntimeException: Stack underrun
	at sootup.java.bytecode.frontend.OperandStack.pop(OperandStack.java:90)
	at sootup.java.bytecode.frontend.OperandStack.popImmediate(OperandStack.java:162)
	at sootup.java.bytecode.frontend.OperandStack.popImmediate(OperandStack.java:172)
	at sootup.java.bytecode.frontend.AsmMethodSource.convertInvokeDynamicInsn(AsmMethodSource.java:1341)
	at sootup.java.bytecode.frontend.AsmMethodSource.convert(AsmMethodSource.java:1759)
	at sootup.java.bytecode.frontend.AsmMethodSource.resolveBody(AsmMethodSource.java:198)
	... 6 more
```

I think the issue lies here https://github.com/soot-oss/SootUp/blob/v1.1.0/sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmMethodSource.java#L1171

My understanding is we shouldn’t pop the operand stack to generate the method handle, as this leaves us with an empty stack when we generate the arguments here https://github.com/soot-oss/SootUp/blob/v1.1.0/sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmMethodSource.java#L1354. 

I’ve currently been working around this by adding a dummy “unknown value”, as opposed to popping the operand stack. 

Aside from that, I think there some logic issues in this stretch of code: 

https://github.com/soot-oss/SootUp/blob/v1.1.0/sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmMethodSource.java#L1383

The variable ```nrArgs``` is set to the size - 1 to account for arrays starting at 0, however in line 1391 (https://github.com/soot-oss/SootUp/blob/v1.1.0/sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmMethodSource.java#LL1391C1-L1392C1), we preemptively decrease the value of nrArgs by 1 again before entering the while statement, meaning we are never popping the full extent of the operand stack. 

Finally, I think the check if the expression is a static invocation here is backwards (https://github.com/soot-oss/SootUp/blob/v1.1.0/sootup.java.bytecode/src/main/java/sootup/java/bytecode/frontend/AsmMethodSource.java#L1394). 

If the expression is static, we’d like to pop the stack once more, not the other way around. 

Apologies for the convoluted message in advance! Resolving the 2 issues I mentioned above helped me get rid of the error messages I was getting but I’d much rather verify if this is correct. 

Thanks!

Sonia

JonasKlauke:
fixes the stack problem in the convertion of invokedynamics


sootup.java.bytecode/src/test/java/sootup/java/bytecode/minimaltestsuite/java8/MethodAcceptingLamExprTest.java
```diff
@@ -43,7 +43,7 @@ public void test() {
   public List<String> expectedBodyStmts() {
     return Stream.of(
             "l0 := @this: MethodAcceptingLamExpr",
-            "l1 = dynamicinvoke \"calcPercentage\" <Percentage ()>() <java.lang.invoke.LambdaMetafactory: java.lang.invoke.CallSite metafactory(java.lang.invoke.MethodHandles$Lookup,java.lang.String,java.lang.invoke.MethodType,java.lang.invoke.MethodType,java.lang.invoke.MethodHandle,java.lang.invoke.MethodType)>(methodtype: double __METHODTYPE__(double), handle: <MethodAcceptingLamExpr: double lambda$lambdaAsParamMethod$0(double)>, methodtype: double __METHODTYPE__(double))",
+            "l1 = dynamicinvoke \"calcPercentage\" <Percentage ()>() <java.lang.invoke.LambdaMetafactory: java.lang.invoke.CallSite metafactory(java.lang.invoke.MethodHandles$Lookup,java.lang.String,java.lang.invoke.MethodType,java.lang.invoke.MethodType,java.lang.invoke.MethodHandle,java.lang.invoke.MethodType)>(methodtype: double __METHODTYPE__(double), methodhandle: \"REF_INVOKE_STATIC\" <MethodAcceptingLamExpr: double lambda$lambdaAsParamMethod$0(double)>, methodtype: double __METHODTYPE__(double))",
```
swissiety, line 46: I guess thats how old soot does it?

SoniaZaldana:
Hi folks, I've been running some tests on my end with this patch and I wanted to note I still see some issues with indys. 
For example
```
public Stream<LocalDate> test(LocalDate endExclusive, Period step) {
        long months = step.toTotalMonths();
        long days = step.getDays();
        int sign = months > 0 || days > 0 ? 1 : -1;
        long steps = 200000 / (months + days);
        return LongStream.rangeClosed(0, steps).mapToObj(
                n -> endExclusive.plusDays(days * n));
    }
```

issues the following error:
```
Error: java.lang.RuntimeException: Failed to convert <jlink.Q: java.util.stream.Stream test(java.time.LocalDate,java.time.Period)>
```

Printing the stack trace
```
java.lang.RuntimeException: Illegal Assignment statement. Make sure that right hand side has a valid operand.
	at sootup.core@1.1.2-SNAPSHOT/sootup.core.jimple.common.stmt.JAssignStmt.<init>(JAssignStmt.java:55)
	at sootup.core@1.1.2-SNAPSHOT/sootup.core.jimple.Jimple.newAssignStmt(Jimple.java:530)
	at sootup.java.bytecode@1.1.2-SNAPSHOT/sootup.java.bytecode.frontend.StackFrame.mergeIn(StackFrame.java:160)
	at sootup.java.bytecode@1.1.2-SNAPSHOT/sootup.java.bytecode.frontend.AsmMethodSource.convertInvokeDynamicInsn(AsmMethodSource.java:1402)
	at sootup.java.bytecode@1.1.2-SNAPSHOT/sootup.java.bytecode.frontend.AsmMethodSource.convert(AsmMethodSource.java:1774)
	at sootup.java.bytecode@1.1.2-SNAPSHOT/sootup.java.bytecode.frontend.AsmMethodSource.resolveBody(AsmMethodSource.java:215)
	at sootup.core@1.1.2-SNAPSHOT/sootup.core.model.SootMethod.lazyBodyInitializer(SootMethod.java:98)
	at com.google.common@32.1.2-jre/com.google.common.base.Suppliers$NonSerializableMemoizingSupplier.get(Suppliers.java:181)
	at sootup.core@1.1.2-SNAPSHOT/sootup.core.model.SootMethod.getBody(SootMethod.java:177)
```

I haven't had a chance to do much digging to see if this is still somehow related to the operand stack but I thought I would bring it up. Thanks!

swissiety:
With Sonias example input there was `null` on the rhs of the assigment.
Then I merged the other porting regression bug fix where we missed handling one Operand.
And now we have:
a) "Invalid in operands length! " in StackFrame with the example

Bildschirmfoto von 2023-08-14 23-00-03
seems the static invoke itself is missing from before..?

b) SwitchExprWithYieldTest fails due to a stack underrun

so it seems there are missing operands that are not added.. or we are taking too much.

sootup.java.bytecode/src/test/java/sootup/java/bytecode/minimaltestsuite/java8/MethodAcceptingLamExprTest.java
JonasKlauke: Exactly. I figured to stay compatible I should just implement it like soot

JonasKlauke:
I have investigated Test: SwitchExprWithYieldTest of java.bytecode.
The stack underrun is probaly caused by handling a virtual invoke by the dynamic invoke, but the invoke should be static. That consumes one element from the stack which is missing for the parameter of the dynamic invoke call.
This seems to only happen if a dynamic invoke is converted for a second time, which is caused by the switch since there are two branches over the same dynamic invoke instruction.

this line is probably the issue:
 final boolean isStaticInvokeExpr = expr instanceof JStaticInvokeExpr;
 since the expr is always a dynamic invoke. the coverage report also shows that the if branch (boolean true) is never entered 

But couldnt find the time to further investigate this, today.

swissiety:
@SoniaZaldana does the current state of this PR fix the indy problems in your project?

swissiety (commented):
:+1: seems reasonable and fixes the submitted failing example input

swissiety (approved):

PR #671, opened 2023-08-18
Update PathBasedAnalysisInputLocation.java

slimming-fat:
If this condition is not added when using SootUp to analyze this Artifact (**org.apache.dubbo:dubbo:3.1.4**),  the following error will be reported. The reason for the error is mistaking the **DENY_CLASS** in org.apache.dubbo:dubbo:3.1.4 as a *.class file.
image
image

### DIFF
--- a/sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/PathBasedAnalysisInputLocation.java
+++ b/sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/PathBasedAnalysisInputLocation.java
@@ -197,6 +197,7 @@ public class PathBasedAnalysisInputLocation implements AnalysisInputLocation<Jav
           .filter(
               filePath ->
                   PathUtils.hasExtension(filePath, handledFileType)
+                  && filePath.toString().endsWith("." + handledFileType.getExtension())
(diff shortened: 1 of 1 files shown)


swissiety:
Thanks for pointing that out - and fixing it!
I'm thinking about whether we should fail here early.. We better catch the Exception (I assume it happens after a call to `view.getClasses(...)`) and skip an invalid file from collecting as well.

JonasKlauke:
Hi thanks for spotting this bug. Nice idea. 
But i think the better fix is to fix hasExtension(..). 
It just checks if the string ends with for example "class" but it should check for ".class"

JonasKlauke:
Nope swisstey, it happens in the fetching of all files in the directory. The classes arent created at that time

swissiety:
> Nope swisstey, it happens in the fetching of all files in the directory. The classes arent created at that time

And why does sootup do that? It is from a call that originates from view.getClasses(..) in a regular usage as described docs ;-)
We are accessing the file contents already (**ClassReader** throws an Exception that this file is malformed).

So we should handle that situation differently in the higher abstraction layer.

JonasKlauke:
This fix was revised in PR #672

Issue #705, opened 2023-10-09
Excpetions in MethodDispatchResolver


Liyw979:
inside `resolveConcreteDispatch`
```
      ClassType finalSuperClassType = superClassType;
      SootClass<?> superClass =
          view.getClass(superClassType)
              .orElseThrow(
                  () ->
                      new ResolveException(
                          "Did not find class " + finalSuperClassType + " in View"));

      classesInHierachyOrder.add(superClass);
```
A `ResolveException` was thrown and it will skip the searching of default methods.
Can I make a PR to remove this throw in `resolveConcreteDispatch`?

swissiety:
this seems reasonable. i.e. not failing completely, working with the information we have (given in the setup).

When you are at it maybe you can simplify the code and replace the calls to findConcreteMethodInSootClass( sootClass, methodSignature ) to sootClass.getMethod(subsignature) as well as they are equivalent.

Liyw979:
nice tip @swissiety

PR #707, opened 2023-10-09
Don't throw execption when class not found in resolveConcreteDispatch

Liyw979:
#705
use `incompleteSuperClassesOf` to prevent class not found

### DIFF
--- a/sootup.core/src/main/java/sootup/core/typehierarchy/MethodDispatchResolver.java
+++ b/sootup.core/src/main/java/sootup/core/typehierarchy/MethodDispatchResolver.java
@@ -184,7 +184,7 @@ public final class MethodDispatchResolver {
     TypeHierarchy hierarchy = view.getTypeHierarchy();
     ClassType superClassType = m.getDeclClassType();
     SootClass<?> startClass = view.getClass(superClassType).orElse(null);
-    ArrayList<SootClass<?>> classesInHierachyOrder = new ArrayList<>();
+    ArrayList<SootClass<?>> classesInHierarchyOrder = new ArrayList<>();
 
     // search concrete method in the class itself and its super classes
     do {
@@ -196,7 +196,7 @@ public final class MethodDispatchResolver {
                       new ResolveException(
                           "Did not find class " + finalSuperClassType + " in View"));
 
-      classesInHierachyOrder.add(superClass);
+      classesInHierarchyOrder.add(superClass);
 
       SootMethod concreteMethod = superClass.getMethod(m.getSubSignature()).orElse(null);
       if (concreteMethod != null && !concreteMethod.isAbstract()) {
@@ -209,7 +209,7 @@ public final class MethodDispatchResolver {
           // A not implemented method of an abstract class results into an abstract method
           return Optional.empty();
         }
-        // found method is abstract and the startclass is not abstract
+        // found method is abstract and the start class is not abstract
         throw new ResolveException(
             "Could not find concrete method for " + m + " because the method is abstract");
       }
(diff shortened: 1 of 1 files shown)


sootup.core/src/main/java/sootup/core/typehierarchy/MethodDispatchResolver.java
```diff
@@ -172,6 +174,19 @@ public static boolean canDispatch(
...
+   * Returns all superclasses of <code>classType</code>(inclusive) up to <code>java.lang.Object</code>, which
+   * will be the last entry in the list, or till one of the superclasses is not contained in view.
+   */
+  private static List<SootClass<?>> findSuperClassesInclusive(
+          View<? extends SootClass<?>> view, ClassType classType) {
+    return Stream.concat(
+                    Stream.of(classType),
+                    view.getTypeHierarchy().incompleteSuperClassesOf(classType).stream()
+            ).flatMap(t -> view.getClass(t).map(Stream::of).orElseGet(Stream::empty))
+            .collect(Collectors.toList());
```
swissiety, line 187: please return the stream and work on it directly - avoiding the copy/collect

sootup.core/src/main/java/sootup/core/typehierarchy/MethodDispatchResolver.java
```diff
@@ -181,23 +196,12 @@ public static boolean canDispatch(
...
+    ClassType current = m.getDeclClassType();
+    SootClass<?> startClass = view.getClass(current).orElse(null);
+    List<SootClass<?>> classesInHierarchyOrder = findSuperClassesInclusive(view, current);
 
-      classesInHierachyOrder.add(superClass);
-
-      SootMethod concreteMethod = findConcreteMethodInSootClass(superClass, m).orElse(null);
+    for (SootClass<?> currentClass : classesInHierarchyOrder) {
+      SootMethod concreteMethod = currentClass.getMethod(m.getSubSignature()).orElse(null);
       if (concreteMethod != null && !concreteMethod.isAbstract()) {
```
swissiety: can you unify this if-else with the following one?

sootup.core/src/main/java/sootup/core/typehierarchy/MethodDispatchResolver.java
```diff
@@ -208,18 +212,16 @@ public static Optional<MethodSignature> resolveConcreteDispatch(
           // A not implemented method of an abstract class results into an abstract method
           return Optional.empty();
         }
-        // found method is abstract and the startclass is not abstract
+        // found method is abstract and the startClass is not abstract
         throw new ResolveException(
             "Could not find concrete method for " + m + " because the method is abstract");
```
swissiety: I think we should ~~log the current ResolveException here and~~ return Optional.empty() as well

Liyw979: I think so, since the method is already returning a Optional.

swissiety: or better change it to IllegalStateException as we should already have found an implemented method for it (as the startClass is not abstract or it is abstract and its superclass has a method that is abstract as well.. so an exception is correct.

swissiety: but when the startclass is already abstract we should return an Optional.empty() the same way as if it would *not* implement an abstract method from its superclass

Liyw979: I think it's better keep it `ResolveException` to stay the same with other exceptions in the file.

sootup.core/src/main/java/sootup/core/typehierarchy/MethodDispatchResolver.java
Liyw979: will do

swissiety (approved):
:+1:

PR #726, opened 2023-10-27
not throw exceptions when class of annotation not found

Liyw979:
When analysing single jar package, Sootup throws exceptions when some class of the annotation is missing.
In this PR:
1. Remove prefetch in `getAnnotations` since `getDefaultValues` is called in `resolveDefaultsForAnnotationType`.
2. Return empty map when the corresponding class is not found.
3. Allow users to do `getAnnotations` without  the default values.


sootup.java.core/src/main/java/sootup/java/core/JavaSootMethod.java
```diff
@@ -68,10 +68,7 @@ public boolean isStaticInitializer() {
 
   @Nonnull
   public Iterable<AnnotationUsage> getAnnotations(@Nonnull Optional<JavaView> view) {
-    annotations.forEach(e -> e.getAnnotation().getDefaultValues(view));
```
JonasKlauke, line 71: @swissiety do you know if this line is important? Is it maybe needed because of the lazy loading?

sootup.java.core/src/main/java/sootup/java/core/types/AnnotationType.java
```diff
@@ -70,12 +64,8 @@ public Map<String, Object> getDefaultValues(@Nonnull Optional<JavaView> viewOpti
         JavaView jv = viewOptional.get();
 
         // meta annotations are not in the view
-        if (this.isMetaAnnotation()) {
-          return defaultValues;
-        }
-
         if (!jv.getClass(this).isPresent()) {
-          throw new RuntimeException("Class of annotation not in view");
```
JonasKlauke, line 78: Pls add a logger warning for the missing annotation

JonasKlauke (changes requested):
Thanks for the small fix. Could you add a warning showing the missing annotation. So User will know that SootUp might need more data.

sootup.java.core/src/main/java/sootup/java/core/JavaSootMethod.java
Liyw979: `getDefaultValues` is called in `resolveDefaultsForAnnotationType`, no need to call it twice.

JonasKlauke (approved):
Thanks 👍

PR #734, opened 2023-10-30
Improve CHA runtime and improve InvokeInterface resolving

### LINKED ISSUE #620: CHA takes a lot of time
CHA seems to take a lot of time. Investigate why it takes so long

### LINKED ISSUE #715: Sources marked as Library shouldn't be further analyzed in the call graph generation
If the sourcetype of an inputlocation is defined as "Library", all methods inside the the location shouldnt be further analyzed if they are reached in the call graph generation.

### LINKED ISSUE #727: Enable to check if an interface method is a default method
Normal interface methods have a body in Bytecode. Therefore, it is not clear if an interface method has a body generated by the compiler or if the body was actually implemented in a default method.

Enable to check if a method is default or set normal interface methods as abstract methods

### LINKED ISSUE #728: The ICFG DotExporter should use a call graph to decide which methods are called
Currently the ICFG Exporter uses the least precise way to resolve call targets. It is not necessary to resolve the call targets again. A call graph should be passed to know what call targets are possible.

JonasKlauke:
- Reworked CHA algorithm to save runtime
- Removed Utilityclass MethodDispatchResolver
- ICFGDotExporter uses callgraph to decide dispatches
- Callgraph algorithm ignore Library classes
- Added Comparator to sort classes by their hierarchy


swissiety (approved):
looks good!
just not sure about whether the other methods (currently unused by the framework) from `MethodDispatchResolver` would still be a nice addition to the framework?

JonasKlauke:
Which Method do you mean?
I think the only interesting method would be concreteDispatch which is still contained in the AbstractCallGraphAlgorithm.
I'm not sure how interesting a method is which just list all implementations of the same method in Subtypes, because subtypes which have no implementation could use default methods of interfaces. 
Additionally, it is quite easy to develop such a method with our provided API
```
MethodSignature baseMethod ....
view.getTypeHierarchy().subtypesOf(classtype)
   .map(clazz ->view.getIdentifierFactory().getMethodSignature(
      clazz , baseMethod.getSubMethodSignature))
   .map(ms -> view.getMethod(ms))
   .filter(Optional::isPresent)
   .map(Optional::get)
```
Sourcecode might not be correct. I wrote this down from mind

JonasKlauke:
But we could think about if we want to provide an option to define a maximum depth for a call graph algorithm. For example a method which just looks at the given method and does not continue further. But for RTA we have the problem that we miss instantiations of classes which result in new targets. But CHA should be okay

PR #738, opened 2023-11-02
fix #682 from path

### LINKED ISSUE #675: JavaIdentifierFactory's fromPath may crash when handling paths with META-INF
code: [https://github.com/soot-oss/SootUp/blob/0fbc9cf0050ef2d38291453e10a5b0f0cdba70ef/sootup.java.core/src/main/java/sootup/java/core/JavaIdentifierFactory.java#L243](https://github.com/soot-oss/SootUp/blob/0fbc9cf0050ef2d38291453e10a5b0f0cdba70ef/sootup.java.core/src/main/java/sootup/java/core/JavaIdentifierFactory.java#L243)

```
    if (path.startsWith("/META-INF/")) {
      // start at 4th separator
      int index = StringUtils.ordinalIndexOf(path, separator, 4);
      path = path.substring(index);
    }
```

StringUtils.ordinalIndexOf directly retrieves the separator's occurrence for the fourth time. If there are only three path separators, it will result in an index of -1, leading to an 'out of range' error when substring is called.

Example path: `/META-INF/quercus/com.caucho.quercus.QuercusClass`

image

The simplest fix:

```
if (path.startsWith("/META-INF/")) {
    // start at 4th separator
    if(StringUtils.countMatches(path, separator) >= 4){
        int index = StringUtils.ordinalIndexOf(path, separator, 4);
        path = path.substring(index);
    }
}
```

However, this fix method is not rigorous. It is still necessary to determine whether it is a multi-release JAR to ensure the correct parsing of data.

### LINKED ISSUE #682: Improve Fully Qualified Name Resolving for ClassTypes
- [x] support multi-release jars which have classes in META-INF
- ~~extract the fully qualified name instead of using the path~~ chicken/egg problem

swissiety:
- [x] check if the requested fq classname matches the fq name in the bytecode
- [x] fix fully qualified Names from `JrtFileSystemAnalysisInputLocation` Module
- [x] fix multi release jar package name generation
~~- [ ] (re)implement multi release for modules~~ -> Issue #823 
- [x] fixed wrong multi-release-jar test input jars
- [x] fix single class file input location ` ClassFileBasedAnalysisInputLocation` as the packagename is not resolved correctly (i.e. missing)
- [x] enable license header checks for all sootup modules; add licenses; add own style check pipeline step which takes care of license headers too

closes #682
fixes #675


sootup.core/src/main/java/sootup/core/IdentifierFactory.java
```diff
@@ -245,18 +244,9 @@ FieldSignature getFieldSignature(
...
-   * @param file the file
-   * @param rootDirectory root directory in which the file is.
-   * @return the class type
-   */
-  ClassType fromPath(Path rootDirectory, Path file);
-
   boolean isStaticInitializerSubSignature(@Nonnull MethodSubSignature methodSubSignature);
 
-  boolean isConstructorSubSignature(@Nonnull MethodSubSignature methodSubSignature);
+  boolean isConstructorSubSignature(@Nonnull MethodSignature methodSignature);
```
stschott: `boolean isConstructorSignature(@Nonnull MethodSignature methodSignature);`

sootup.java.core/src/main/java/sootup/java/core/JavaIdentifierFactory.java
```diff
@@ -56,14 +54,21 @@ public class JavaIdentifierFactory implements IdentifierFactory {
 
   @Nonnull private static final JavaIdentifierFactory INSTANCE = new JavaIdentifierFactory();
 
+  @Nonnull
+  public final MethodSubSignature STATIC_INITIALIZER =
```
stschott: Can't this be a static variable to make it a constant?

sootup.java.core/src/main/java/sootup/java/core/JavaIdentifierFactory.java
```diff
@@ -56,14 +54,21 @@ public class JavaIdentifierFactory implements IdentifierFactory {
...
+      new MethodSubSignature("<clinit>", Collections.emptyList(), VoidType.getInstance());
+
   @Override
   public boolean isStaticInitializerSubSignature(@Nonnull MethodSubSignature methodSubSignature) {
```
stschott, line 62: why not compare it to the above introduced constant ?

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/PathBasedAnalysisInputLocation.java
```diff
@@ -100,84 +111,72 @@ public List<BodyInterceptor> getBodyInterceptors() {
...
+  @Nonnull
+  public static PathBasedAnalysisInputLocation create(
+      @Nonnull Path path,
+      @Nonnull SourceType srcType,
+      @Nonnull List<BodyInterceptor> bodyInterceptors,
+      @Nonnull Collection<Path> ignoredPaths) {
+
+    if (ignoredPaths.stream()
+        .anyMatch(ignoPath -> path.toString().startsWith(ignoPath.toString()))) {
+      throw new IllegalArgumentException(
```
stschott, line 134: Shouldn't ignored input locations just not be considered for analysis?

stschott (approved):

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/PathBasedAnalysisInputLocation.java
swissiety: yes, but this way we can cut the loading earlier - especially it creates problems e.g. MultiReleaseJars would point to the jars root - which contains META-INF/versions/ as a subfolder and therefore would translate the complete path as the package name (and would resolve files it shouldnt); as a sideeffect this enables more precise "scoping" (again) as we removed the DefaultSourceTypeSpecifier

PR #744, opened 2023-11-07
Fix two issues with DominanceFinder

### LINKED ISSUE #586: Dominance Finder Out of Bounds Error
When I attempt to build a DominanceFinder object, I am getting an Array Out of Bounds Exception:

Here is the snippet of my code which produces the error:
```java
for(JavaSootMethod aMethod : aClass.getMethods()){ 
    if(aMethod.hasBody()){
        StmtGraph<?> graph = aMethod.getBody().getStmtGraph();
        df = new DominanceFinder(graph);
        ...
```

Here is the stack trace:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 18
	at sootup.core.graph.DominanceFinder.getFirstDefinedBlockPredIdx(DominanceFinder.java:167)
	at sootup.core.graph.DominanceFinder.<init>(DominanceFinder.java:75)
	at com.lesslab.ProgramGraph.GraphBuilder.buildGraphs(GraphBuilder.java:193)
	at com.lesslab.ProgramGraph.App.main(App.java:45)
```

Liam-DeVoe:
First off: apologies for bundling two fixes with this PR. I would have submitted them separately, but the test I added runs into both issues, and so would fail on any PR which included only a fix for one. I've done the next best thing and split the fixes into commits (c07888eea3ca1f6e3e2a5b50b0fe5b0f2235f43d and b5058575ac1657375370362ef724cf6e7e0653fc). However, I'm happy to split these into PRs on request.

### c07888eea3ca1f6e3e2a5b50b0fe5b0f2235f43d: use .equals to compare to starting stmt

This closes https://github.com/soot-oss/SootUp/issues/586 by using `.equals` for comparison instead of `==`.

### b5058575ac1657375370362ef724cf6e7e0653fc: use correct reverse postorder iteration for dominance finder

While investigating https://github.com/soot-oss/SootUp/issues/586, I found that `DominanceFinder` was not computing the correct dominators for the test I added. I think the reasons for this are two-fold:

* To start, on master, `DominanceFinder#getIdxToBlock` and `DominanceFinder#getBlockToIdx` are *not* inverses (`git checkout c07888eea` and running `testBlockToIdxInverse` confirms this).  This causes dominance indexes to become mixed up internally and the wrong dominator to be returned.
* Even if this is fixed in a simple way, I don't believe that iterating over `blockGraph.getBlocks()` to compute immediate dominators is correct. The [algorithm referenced](https://www.cs.tufts.edu/comp/150FP/archive/keith-cooper/dom14.pdf) requires that blocks are iterated over in reverse postorder. `blockGraph.getBlocks()`'s xmldoc indicates that its ordering is not only not reverse postorder, but is not necessarily in any deterministic order.

I've chosen to iterate over `blockGraph.getBlockIterator()` instead, which fixes the first issue by using `blockGraph.getBlockIterator()` as the ordered source of truth for nodes (and building the backwards `blockToIdx` map straight from it). This also fixes the second issue, as `blockGraph.getBlockIterator()` seems to provide the nodes in reverse postorder, as required. However, I'm not sure if this is an invariant I can rely on - let me know if it's not.

I also fixed a dead pdf link along the way.

### DIFF
--- a/sootup.core/src/main/java/sootup/core/graph/DominanceFinder.java
+++ b/sootup.core/src/main/java/sootup/core/graph/DominanceFinder.java
@@ -30,7 +30,8 @@ import javax.annotation.Nonnull;
 /**
  * @author Zun Wang
  * @see <a
- *     href="https://www.cs.rice.edu/~keith/EMBED/dom.pdf">https://www.cs.rice.edu/~keith/EMBED/dom.pdf</a>
+ *     href="https://www.researchgate.net/publication/2569680_A_Simple_Fast_Dominance_Algorithm">
+ *     https://www.researchgate.net/publication/2569680_A_Simple_Fast_Dominance_Algorithm </a>
  */
 public class DominanceFinder {
 
--- a/sootup.core/src/main/java/sootup/core/graph/DominanceFinder.java
+++ b/sootup.core/src/main/java/sootup/core/graph/DominanceFinder.java
@@ -23,6 +23,8 @@ package sootup.core.graph;
  */
 
 import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 import javax.annotation.Nonnull;
 
 /**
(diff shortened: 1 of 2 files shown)


sootup.core/src/main/java/sootup/core/graph/DominanceFinder.java
```diff
@@ -39,19 +42,22 @@ public class DominanceFinder {
 
   public DominanceFinder(StmtGraph<?> blockGraph) {
 
-    // assign each block a integer id, startBlock's id must be 0
-    blocks = new ArrayList<>(blockGraph.getBlocks());
+    // we're locked into providing a List<BasicBlock<?>>, not a List<? extends BasicBlock<?>>, so
+    // we'll use the block iterator directly (which provides this type) rather than
+    // #getBlocksSorted.
+    blocks =
```
Liam-DeVoe, line 48: as mentioned in the comment, this seems like an unfortunate api clash between what `blockGraph.getBlocksSorted()` returns (`List<? extends BasicBlock<?>>`) and what `DominanceFinder#getIdxToBlock` returns (`List<BasicBlock<?>>`). 

I didn't want to make a breaking api change in this pr (though if you're okay with it, that might be a good alternative!), so I've used `blockGraph.getBlockIterator()` here instead, which has the desired return type.

kadirayk (approved):
Thanks a lot! 👍

PR #777, opened 2023-12-17
New API: get method by name

Liyw979:


JonasKlauke:
mmh I think getMethodByName would be a better name otherwise you might think you can input a string of a method signature

sootup.core/src/main/java/sootup/core/model/AbstractClass.java
```diff
@@ -122,4 +123,15 @@ public Optional<? extends SootMethod> getMethod(
                   getClassSource().getSourcePath());
             });
   }
+
+  /**
+   * Attempts to retrieve the method with the given name. This method will return an empty Set if
+   * there is no method with the given name.
```
JonasKlauke, line 129:
```suggestion
   * Attempts to retrieve the method with the given name. This method will return an empty Set if
   * there is no method with the given name.
   * @param name the name of the method
   * @return a set of methods that have the given name
```

Liyw979:
> mmh I think getMethodByName would be a better name otherwise you might think you can input a string of a method signature

I think so.

JonasKlauke (approved):
👍

PR #782, opened 2023-12-19
replace throw with log and empty set

Liyw979:


swissiety:
I would not remove these exceptions as they mean: the classtype is not in the typehierarchy vs. classtype has no e.g. subclasses - would change them to sth like NoSuchElementException due to [this reasoning](https://stackoverflow.com/a/77361) - maybe even as checked exception.
(“If a client can reasonably be expected to recover from an exception, make it a checked exception. If a client cannot do anything to recover from the exception, make it an unchecked exception.”)

JonasKlauke:
If theses methods cause problems in SootUp itself, these areas should catch the expections and print a warning before they continue with the incomplete data. I agree that the exceptions should not removed since we can not differentiate between missing data and the result that there is nothing

PR #783, opened 2023-12-21
Fix CallGraphAlgorithm.addClass(..) implementation

swissiety:
- [x] in case of non-existing methods in the CG it threw:

```
Exception in thread "main" java.lang.NullPointerException: Node for <org.apache.poi.ss.formula.functions.FreeRefFunction: org.apache.poi.ss.formula.eval.ValueEval evaluate(org.apache.poi.ss.formula.eval.ValueEval[],org.apache.poi.ss.formula.OperationEvaluationContext)> has not been added yet
```

- [x] don't fail if nothing new was added (or a subtask was already calculated before)
- [x] return more convenient Signature from JavaSootClass.getType()

### DIFF
--- a/sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
+++ b/sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
@@ -319,7 +319,6 @@ public abstract class AbstractCallGraphAlgorithm implements CallGraphAlgorithm {
     Set<MethodSignature> newMethodSignatures =
         clazz.getMethods().stream().map(Method::getSignature).collect(Collectors.toSet());
 
-
     if (newMethodSignatures.stream().anyMatch(oldCallGraph::containsMethod)) {
       // FIXME: [ms] handle better - remove from entry point signatures in this case
       throw new IllegalArgumentException("CallGraph already contains methods from " + classType);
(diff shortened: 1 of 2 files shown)


kadirayk (approved):

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -354,8 +359,10 @@ public CallGraph addClass(@Nonnull CallGraph oldCallGraph, @Nonnull JavaClassTyp
               MethodSignature overridingMethodSig =
                   clazz.getMethod(overriddenMethodSig.getSubSignature()).get().getSignature();
 
-              for (MethodSignature callingMethodSig : updated.callsTo(overriddenMethodSig)) {
-                updated.addCall(callingMethodSig, overridingMethodSig);
+              if (updated.containsMethod(overriddenMethodSig)) {
+                for (MethodSignature callingMethodSig : updated.callsTo(overriddenMethodSig)) {
+                  updated.addCall(callingMethodSig, overridingMethodSig);
```
JonasKlauke, line 364: This is only true for CHA algorithm or for the RTA (if the given classtype is instantiated)
If we want to actually support this function the abstaractcallgraphalgorithm has to be abstract and this method has to be developed for each algorithm.

additionally if this class was already in the view all edges are already included.
I think the only reason for this method if you want to force the call graph to contain the class, even when the algorithm would say it is not contained, and then we could do it like this (like CHA)

PR #793, opened 2024-01-03
fix the print of field signatures

stschott:
fixes missed escaping + StmtPrinter.typeSignature() to adapt type printing

### DIFF
--- a/sootup.core/src/main/java/sootup/core/util/printer/JimplePrinter.java
+++ b/sootup.core/src/main/java/sootup/core/util/printer/JimplePrinter.java
@@ -26,16 +26,11 @@ import java.io.PrintWriter;
 import java.util.*;
 import java.util.stream.Collectors;
 import sootup.core.graph.StmtGraph;
+import sootup.core.jimple.Jimple;
 import sootup.core.jimple.basic.Local;
 import sootup.core.jimple.basic.Trap;
 import sootup.core.jimple.common.stmt.Stmt;
-import sootup.core.model.Body;
-import sootup.core.model.ClassModifier;
-import sootup.core.model.Field;
-import sootup.core.model.Method;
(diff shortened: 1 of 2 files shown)


sootup.core/src/main/java/sootup/core/util/printer/JimplePrinter.java
```diff
@@ -196,8 +191,12 @@ public void printTo(SootClass<?> cl, PrintWriter out, LabeledStmtPrinter printer
           SootField f = (SootField) fieldIt.next();
           printer.newline();
           printer.handleIndent();
-          printer.literal(f.getDeclaration());
```
swissiety, line 199: do we need/use/want .getDeclaration() somewhere else? maybe we can remove it.. I mean its nice to have when needed - but I would not expect to find it and its basically just concatenating two SootField methods which I would do quickly myself before searching for getDeclaration() unless I stumbled over it already.

sootup.core/src/main/java/sootup/core/util/printer/JimplePrinter.java
```diff
@@ -196,8 +191,12 @@ public void printTo(SootClass<?> cl, PrintWriter out, LabeledStmtPrinter printer
...
-          printer.literal(";");
+          if (!f.getModifiers().isEmpty()) {
+            printer.literal(FieldModifier.toString(f.getModifiers()));
```
swissiety, line 195: Offtopic: FieldModifier.toString(...) already assumes valid Modifier combinations (theoretically its currently possible to assign public+private+protected modifier at the same time) - maybe we should implement a wrapper of EnumSet<...Modifier> to handle only valid combinations and implement toString there?

swissiety (approved):

PR #818, opened 2024-01-14
`HasPosition` Interface

DavidBakerEffendi:
In the old Soot, there is a `soot.tagkit.Host` interface which defined methods such as `getJavaSourceStartLineNumber`. Many core classes representing code objects implemented this and allowed libraries implementing Soot to easily define a single method to retrieve position information from various classes.

Such an interface is not present in SootUp, and so this PR presents the `sootup.core.model.HasPosition` interface to define what many classes already have as `getPosition`. All classes that already define `getPosition` now implement `HasPosition` as a part of this change.


swissiety:
sounds reasonable - e.g. for IDE (read: editor) / UI / LSP tooling this would make sense

DavidBakerEffendi:
@swissiety This is largely motivated by Soot's use in Joern. I'm slowly moving over to SootUp: https://github.com/joernio/joern/issues/4054

swissiety (approved):
:+1:

PR #819, opened 2024-01-15
Adds repository declaration to installation docs

rzo1:
Hi,

if one follows the installation procedere described on https://soot-oss.github.io/SootUp/v1.1.2/installation/ with a plain and default Maven installation (no custom artifactory / repository via settings.xml, etc.), several dependencies cannot be retrieved resulting in a build failure.

I suggest to add the required declarations to the documentation, so people have a chance to just consume **or** configure there local artifactory/repo proxies accordingly.


swissiety (approved):
:+1:

PR #821, opened 2024-01-17
fix filesystem caching issues

### LINKED ISSUE #527: Class files are closed and later used
image

### LINKED ISSUE #617: Possibly better cache strategy for FileSystem
https://github.com/soot-oss/SootUp/blob/8ac5289de90f2d68a604930d5d818dce15d2eb88/sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/PathBasedAnalysisInputLocation.java#L551-L575

I think it is better to use reference-based eviction instead of timed eviction, that is `CacheBuilder.weakValues()`.

Firstly, in this way, we can use cached FileSystem in:
https://github.com/soot-oss/SootUp/blob/8ac5289de90f2d68a604930d5d818dce15d2eb88/sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/PathBasedAnalysisInputLocation.java#L600-L608

Secondly, `MultiReleaseJarAnalysisInputLocation.discoverInputLocations` adds archiveRoot to inputLocations:
https://github.com/soot-oss/SootUp/blob/8ac5289de90f2d68a604930d5d818dce15d2eb88/sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/PathBasedAnalysisInputLocation.java#L307
This archiveRoot is a DirectoryBasedAnalysisInputLocation. It uses its `path`'s FileSystem in `getClassSources`, `getClassSource` and some other methods. If the FileSystem is used after closed (1 second timeout), it will cause a ClosedFileSystemException.

swissiety:
no timing based cache eviction (+ resource closing) of a filesystem anymore.

- [ ] check when/if cache entries are actually evicted - eventually closing the fs resource. 
e.g. Path can reference a Filesystem and therefore keep the filesystem open/ the cache entry alive

fixes #527
fixes #617

### DIFF
--- a/sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/ArchiveBasedAnalysisInputLocation.java
+++ b/sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/ArchiveBasedAnalysisInputLocation.java
@@ -86,28 +86,26 @@ public class ArchiveBasedAnalysisInputLocation extends PathBasedAnalysisInputLoc
   @Override
   @Nonnull
   public Optional<JavaSootClassSource> getClassSource(@Nonnull ClassType type, @Nonnull View view) {
-    try (FileSystem fs = fileSystemCache.get(path)){
+      try{
+      FileSystem fs = fileSystemCache.get(path);
       final Path archiveRoot = fs.getPath("/");
       return getClassSourceInternal(
           (JavaClassType) type, archiveRoot, new AsmJavaClassProvider(view));
     } catch (ExecutionException e) {
       throw new RuntimeException("Failed to retrieve file system from cache for " + path, e);
-    } catch (IOException e) {
-        throw new RuntimeException(e);
     }
   }
 
   @Override
   @Nonnull
   public Collection<JavaSootClassSource> getClassSources(@Nonnull View view) {
-    try (FileSystem fs = fileSystemCache.get(path)) {
+    try{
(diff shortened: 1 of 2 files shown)


sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/ArchiveBasedAnalysisInputLocation.java
```diff
@@ -98,14 +98,13 @@ public Optional<JavaSootClassSource> getClassSource(@Nonnull ClassType type, @No
   @Override
   @Nonnull
   public Collection<JavaSootClassSource> getClassSources(@Nonnull View view) {
-    // we don't use the filesystem cache here as it could close the filesystem after the timeout
-    // while we are still iterating
-    try (FileSystem fs = FileSystems.newFileSystem(path, (ClassLoader) null)) {
+    try {
+      FileSystem fs = fileSystemCache.get(path);
```
stschott, line 102: why not keep it in the try-with-resources statement?

stschott (approved):

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/ArchiveBasedAnalysisInputLocation.java
swissiety: filling/using the cache with living entries

PR #825, opened 2024-01-22
fix TypeResolver using field references incorrectly

### LINKED ISSUE #799: fix TypeAssigner
1) lab3:fieldSQLInjection1 - it does not type precisely - but creates casts for all further usages of $stack8

```
    $stack8 = new target.hidden.exercise2and3.FieldSQLInjection1$ObjectWithTaint;
```

2) does not type single defs - but casts them - so the information seems to be gathered

3) splits a single assignment into 2 varibles (maybe connected with 2) ?)

Timbals:
Assignments to fields of locals shouldn't change the types of the locals themselves.

This was bug was introduced in [this commit](https://github.com/soot-oss/SootUp/commit/ebef825474771eb0e8adb976e007050aaba9771c).

Closes #799


JonasKlauke (approved):
Is this done?
looking good

Timbals:
> Is this done?

yes

PR #716, opened 2023-10-25
investigate issue714

swissiety:
starter base - feel free to investigate further!
fix slow_localSplitter():
- [ ] can not find the Method: compares JavaClassType with JavaModuleClassType refering to java.lang.String and fails
- [ ] JavaModuleView.getClasses() only returns classes that are alredy queried by getClass( type )

bloated_AsmMethodSource_worklist:
- [ ] fix view as above to retrieve the method
- [ ] ... 

TypeAssigner not terminating was fixed in the meantime.

closes #714

PR #841, opened 2024-02-07
Rewrite `LocalSplitter`

### LINKED ISSUE #798: Fix LocalSplitter
Overcomplicated code. Infinite Loops...

Timbals:
This PR rewrites the `LocalSplitter` to use a simpler algorithm.

The algorithm uses a [disjoint set](https://en.wikipedia.org/wiki/Disjoint-set_data_structure) to find any definitions for locals that can't be split because they have overlapping uses.
Then it splits the locals by modifying the statements in the graph.

I also did some easy performance optimization. Running the `LocalSplitter` over the entire Java Runtime Library (roughly 200_000 methods of varying length and complexity) takes around 10 seconds on my machine. That seems adequate for now.
There is definitely more room for optimization here (potential for caching results, reduce the amount of hashing done, reduce the amount of expensive `indexOf` calls).

Closes #798
Fixes the problems with the LocalSplitter from #716


swissiety (approved):
good job!

PR #846, opened 2024-02-08
fix cast error for `Operand`s with side effects

Timbals:
`Operand`s that can be used as both a statement and an expression could throw a cast exception. This was not noticed earlier because `AsmMethodSource::getStmt` is generic and therefore no actual cast was visible in the source code.

I also added a testcase for this behavior.

(This was originally discovered while working on #841 in the method `<java.net.URI: java.lang.String decode(java.lang.String,boolean)>`)


swissiety (approved):
Nice. Thx for your contribution!

PR #847, opened 2024-02-08
set BytecodeBodyInterceptors as default for BytecodeFrontend based InputLocations

swissiety:
- [x] refactor ClassLoadingOptions away
- [x] set default BytecodeInterceptors
- [x] fix duplicate predecessorBlocks in StmtGraph
- [x] fix StmtGraph.removeNode() -> keep correct/all predecessor information of a deleted block to the new succesor
- [x] fix ~~the Interceptor that does not update the Set of Locals when it removes Stmts that eliminate Local defs&usages (The TypeAssigner has Locals that do not exist~~ DeadassignmentEliminator: remove Locals, too (**investigate** TypeAssigner did weird with that wrong info and failed!)
- [x] adapt testcases that were designed for use without interceptors to not use the default interceptors 

tried to extract these to keep the  changed files smaller.. unfortunately lead to more errors that are already fixed here.
- [x] check if every Test class has a Category - mentioned in https://github.com/soot-oss/SootUp/issues/849
used regex: (?<!@Category\(Java[8|9]Test\.class\))(\spublic\sclass\s(\S*)[T|t]est) scope: test files
- [x] fix LocalGenerator
- [x] check default interceptors on the rt.jar - fixing them iteratively; fix the failing ones in another PR


swissiety:
~~something is weird~~ - log says: "sootup.java.bytecode.frontend.AsmMethodSourceTest Error:  testNestedMethodCalls(sootup.java.bytecode.frontend.AsmMethodSourceTest)" but the test should not be available in the check run of [3ace299](https://github.com/soot-oss/SootUp/pull/847/commits/3ace299c7cf5d8abf657bbcbad8d8dc80dd3082c) i.e. before I merged develop in which the testcase already exists?

--> the pipeline runs on a merged branch of targetBranch + currentBranch

sootup.core/src/main/java/sootup/core/jimple/common/stmt/AbstractStmt.java
```diff
@@ -145,7 +145,12 @@ public StmtPositionInfo getPositionInfo() {
   @Nullable
   public Stmt withNewUse(@Nonnull Value oldUse, @Nonnull Value newUse) {
     ReplaceUseStmtVisitor visitor = new ReplaceUseStmtVisitor(oldUse, newUse);
-    accept(visitor);
+    try {
+      accept(visitor);
+    } catch (ClassCastException cce) {
+      // new Stmt is not created as the newUse could not be replaced
+      return null;
```
JonasKlauke, line 152: how about optional instead of a possible null as return

sootup.core/src/main/java/sootup/core/jimple/visitor/ReplaceUseStmtVisitor.java
```diff
@@ -96,10 +96,14 @@ public void caseAssignStmt(@Nonnull JAssignStmt stmt) {
...
-        if (refVisitor.getResult() != rValue) {
-          setResult(stmt.withRValue(refVisitor.getResult()));
+        try {
+          refVisitor.init(oldUse, newUse);
+          ((Ref) rValue).accept(refVisitor);
+          if (refVisitor.getResult() != rValue) {
+            setResult(stmt.withRValue(refVisitor.getResult()));
+          }
+        } catch (ClassCastException cce) {
+          // can not replace that local by another Value
```
JonasKlauke, line 106: empty catch block

JonasKlauke: log a warning?

sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/Aggregator.java
```diff
@@ -191,22 +191,24 @@ public void interceptBody(@Nonnull Body.BodyBuilder builder, @Nonnull View view)
...
         final ReplaceUseStmtVisitor replaceVisitor = new ReplaceUseStmtVisitor(val, aggregatee);
-        // FIXME: this try-catch is an awful hack for "ValueBox.canContainValue" -> try to determine
+        // TODO: this try-catch is an awful hack for "ValueBox.canContainValue" -> try to determine
         // a replaceability earlier!
         try {
           replaceVisitor.caseAssignStmt(assignStmt);
           newStmt = replaceVisitor.getResult();
         } catch (ClassCastException iae) {
-          newStmt = null;
+          continue;
```
JonasKlauke, line 200: log warning?

sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/ConditionalBranchFolder.java
```diff
@@ -58,48 +58,78 @@ public void interceptBody(@Nonnull Body.BodyBuilder builder, @Nonnull View view)
...
       // check for constant-valued conditions
       Constant evaluatedCondition = Evaluator.getConstantValueOf(ifStmt.getCondition());
-      if (evaluatedCondition == null) {
+
+      boolean removeTrueBranch;
+      if (evaluatedCondition instanceof BooleanConstant) {
+        removeTrueBranch = evaluatedCondition == BooleanConstant.getTrue();
+      } else if (evaluatedCondition instanceof IntConstant) {
+        removeTrueBranch =
+            IntConstant.getInstance(0).equalEqual(((IntConstant) evaluatedCondition))
```
JonasKlauke, line 67: is instance(0) always set?

sootup.java.bytecode/src/test/java/sootup/java/bytecode/minimaltestsuite/MinimalBytecodeTestSuiteBase.java
```diff
@@ -44,18 +46,22 @@ public static class CustomTestWatcher extends TestWatcher {
     /** Load View once for each test directory */
     @Override
     protected void starting(Description description) {
+      // TODO: seems to be more complicated than necessary - save directory instead of needing
```
JonasKlauke, line 49: i fixed this in the migration 😄

JonasKlauke:
Did this change the public api? If not should we release a 1.2.1 with this

swissiety:
> Did this change the public api? If not should we release a 1.2.1 with this

it should not break from the API from a compilation side -( it was basically an extension/overloading of constructors + fixing stuff in this journey) - but the intention of applying interceptors by default now should be a new minor++ release

sootup.core/src/main/java/sootup/core/jimple/visitor/ReplaceUseStmtVisitor.java
swissiety: we don't have to log as we use this try catch as a means to "is it possible to apply this new value" and if it fails it sets no new result, which conforms to the other uses there

sootup.core/src/main/java/sootup/core/jimple/common/stmt/AbstractStmt.java
swissiety: We could do that.. but would be structurally different to the other `with()`-ers

sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/Aggregator.java
swissiety: nope - same as above (hack to not existing  "isReplaceableByThisValue()")

sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/ConditionalBranchFolder.java
swissiety: never null

JonasKlauke (approved):

PR #856, opened 2024-02-15
Migrate test system from junit 4 to junit 5

JonasKlauke:
Update all tests to the API changes in juint 5


swissiety (changes requested):
good work! 

To avoid `@Tag` problems with typos lets replace `@Tag\(\"Java([8|9])\"\)` by `@Tag(TestCategories.JAVA_$i_CATEGORY)` and add a categories class file to each module like its done in the bytecode module ( or can we move it to core and reuse it in the other modules?)

swissiety:
just from the added timing in the java8 run (+2minutes) I would assume the testwatcher is not caching?
- [x] investigate

JonasKlauke:
I added on many testclasses a Tag which had no category before. That could be the reason. Since we also gained coverage

JonasKlauke:
The Enum Tag is good and yeah we should go for only one Enum. If every module imports core then we can simply put it there

JonasKlauke:
Enum moved to #859

swissiety (approved):

PR #858, opened 2024-02-19
Improve/interceptors

swissiety:
- remove deprecated builder.*stmtgraph-manipulations* - just do it directly 
- readability
- simplify

### DIFF
--- a/sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/ConstantPropagatorAndFolder.java
+++ b/sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/ConstantPropagatorAndFolder.java
@@ -24,6 +24,8 @@ import com.google.common.collect.Lists;
 import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nonnull;
+
+import sootup.core.graph.MutableStmtGraph;
 import sootup.core.jimple.basic.Immediate;
 import sootup.core.jimple.basic.Local;
 import sootup.core.jimple.basic.Value;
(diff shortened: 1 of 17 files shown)


sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/ConstantPropagatorAndFolder.java
```diff
@@ -54,9 +57,12 @@ public void interceptBody(@Nonnull Body.BodyBuilder builder, @Nonnull View view)
 
     // Perform a constant/local propagation pass
     // go through each use in each statement
-    for (Stmt stmt : Lists.newArrayList(builder.getStmts())) {
+    MutableStmtGraph stmtGraph = builder.getStmtGraph();
+    for (Stmt stmt : Lists.newArrayList(stmtGraph)) {
```
JonasKlauke, line 61: why dont you use the normal stmt iterator?

sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/ConstantPropagatorAndFolder.java
```diff
@@ -54,9 +57,12 @@ public void interceptBody(@Nonnull Body.BodyBuilder builder, @Nonnull View view)
...
       // propagation pass
-      if (stmt instanceof JAssignStmt) {
+      boolean isAssignStmt = stmt instanceof JAssignStmt;
+      boolean isReturnStmt = stmt instanceof JReturnStmt;
```
JonasKlauke: both are used only in the if header, why did you create a local variable?

JonasKlauke (approved):

sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/ConstantPropagatorAndFolder.java
swissiety: because we modify the stmtgraph and without the copy it would throw a nice ConcurrentModificationException (before the List was copied twice)

JonasKlauke (approved):

sootup.java.bytecode/src/main/java/sootup/java/bytecode/interceptors/ConstantPropagatorAndFolder.java
swissiety: it was used twice and then after refactoring not needed anymore - will fix it

PR #860, opened 2024-02-19
Method signature validator

akshitad11:
Method Signature Validator Changes


swissiety:
hint: #863 targets the same Validator - but mostly checks for allowed return types

kadirayk (approved):

PR #863, opened 2024-02-21
Adapt Class validators to SootUp

Momo-Not-Emo:
This pull request modifies the `MethodDeclarationValidator` and `ClassFlagsValidator` classes to be compatible with SootUp. 

- The changes include the addition of a `boolean isAllowedInFinalCode()` method to the Type class and its subclasses for `MethodDeclarationValidator`. 
- Additionally, `sc.isSuper()` is now used in `ClassFlagsValidator` instead of `sc.isSynchronized()` due to a method renaming.

However, the adaptation of `OuterClassValidator` remains incomplete. This validator is designed to check for cycles in the outer class chain, but adapting it is challenging because the return type of `SootClass::getOuterClass` has changed from `SootClass` to `ClassType`. I am uncertain about **how to obtain the `SootClass` object from the `ClassType`**.

Note: There are no test cases included in this pull request, as these class validators are not currently utilized in SootUp. They are typically invoked through `PackManager` in Soot, but it appears that SootUp has not yet ported `PackManager`.

### DIFF
--- a/sootup.core/src/main/java/sootup/core/types/ArrayType.java
+++ b/sootup.core/src/main/java/sootup/core/types/ArrayType.java
@@ -105,4 +105,9 @@ public class ArrayType extends ReferenceType {
   public void accept(@Nonnull TypeVisitor v) {
     v.caseArrayType();
   }
+
+  @Override
+  public boolean isAllowedInFinalCode() {
+    return true;
+  }
 }
--- a/sootup.core/src/main/java/sootup/core/types/PrimitiveType.java
+++ b/sootup.core/src/main/java/sootup/core/types/PrimitiveType.java
@@ -111,6 +111,11 @@ public abstract class PrimitiveType extends Type {
     public void accept(@Nonnull TypeVisitor v) {
       v.caseByteType();
     }
+
+    @Override
+    public boolean isAllowedInFinalCode() {
+      return true;
+    }
   }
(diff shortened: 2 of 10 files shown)


swissiety:
thx for your contribution! It seems like a duplicate of #860 ? Lets merge the best of both approaches.

> how to obtain the SootClass object from the ClassType

you can get it via `view.getClass( classType ).get()`

I think isAllowedInFinalCode() should not be necessary (and thats the reason we have/need a validator) / I am not sure we should have an API for it. Maybe lets have an interface for the all allowed types and check via instanceof that interface? like an interface for non-void types, "Method returnable types" including void?

akshitad11 (commented):
Kindly refer to my pull request : [860](https://github.com/soot-oss/SootUp/pull/860).
I have already made changes to MethodDeclarationValidator in much more detail there. Let's merge these changes as suggested by @swissiety.  
Also, my changes for ClassFlagsValidator(now renamed to ClassModifierValidator) are merged in develop, please take the latest version of develop and you will see the changes.

Momo-Not-Emo:
> > how to obtain the SootClass object from the ClassType
> 
> you can get it via `view.getClass( classType ).get()`

However, it seems `view` is not available in the validator. Should it be passed as a parameter of `void validate(...)`?

PR #865, opened 2024-02-21
InvokeArgumentValidator changes

Momo-Not-Emo:
1. Adapt InvokeArgumentValidator and add test cases. The original version primarily focuses on verifying the number of arguments. 
2. Implement argument type checking by including `rt.jar` in the analysis input locations, as the validator relies on typeHierarchy. Add a corresponding test case for this scenario.
3. Rectify a bug in `JavaIdentifierFactory.getBoxedType(...)` related to the int type. Replace the erroneous `java.lang.Int` with the correct type, which is `java.lang.Integer`.
4. Add support to check implicit conversion for primitive types.
5. Add a workaround check in `ArchiveBasedAnalysisInputLocation.getClassSource(...)` to avoid casting exception when passing an anonymous `ClassType` object which cannot be casted to `JavaClassType` directly.


sootup.core/src/main/java/sootup/core/validation/InvokeArgumentValidator.java
```diff
@@ -22,31 +22,42 @@
...
-     */
-    return null;
-  }
+        for (Stmt stmt : body.getStmts()) {
+            if (stmt.containsInvokeExpr()) {
+                AbstractInvokeExpr invExpr =
+                        stmt.getInvokeExpr();
+                MethodSignature callee = invExpr.getMethodSignature();
+                if (invExpr.getArgCount() !=
+                        callee.getParameterTypes().size()) {
```
swissiety: This condition is always false as they point to the same data - we need to get a SootMethod via the view.

sootup.core/src/main/java/sootup/core/validation/InvokeArgumentValidator.java
```diff
@@ -22,31 +22,42 @@
...
+                    validationException.add(new ValidationException(stmt, "Invalid number of arguments"));
+                }
+            }
```
swissiety, line 88: An additional check for correct argument types (+inheritance - use the view.typehierarchy() ) would be nice, too.

sootup.tests/src/test/java/sootup/tests/validator/InvokeArgumentValidatorTest.java
```diff
@@ -0,0 +1,114 @@
...
+                new JimpleAnalysisInputLocation(Paths.get(classPath), SourceType.Application);
+
+        jimpleView = new JimpleView(jimpleInputLocation);
+        final Optional<SootClass> classSource1 = jimpleView.getClass(classTypeFieldRefValidator);
+        assertFalse(classSource1.isPresent());
+
+        classes = new HashSet<>(); // Set to track the classes to check
+
+        for (SootClass aClass : jimpleView.getClasses()) {
+            if (!aClass.isLibraryClass()) {
```
swissiety: isLibraryClass check is not necessary as only an AnalysisInputLocation with SourceType.Application is added to the view

sootup.tests/src/test/java/sootup/tests/validator/InvokeArgumentValidatorTest.java
```diff
@@ -0,0 +1,114 @@
...
+                        classes.stream()
+                                .filter(c -> c.getType().getClassName().equals("InvokeArgumentValidator"))
+                                .findFirst()
+                                .get()
+                                .getMethods()
+                                .stream()
+                                .filter(m -> m.getName().equals("invokeArgumentValidator_success"))
+                                .map(m -> m.getBody())
+                                .findFirst()
+                                .get(),
```
swissiety: you can simplify accessing the methodBody: `view.getMethod( view.getIdentifierFactory().parseMethod("<InvokeArgumentValidator: void invokeArgumentValidator_success()>") ).get().getBody()`

sootup.core/src/main/java/sootup/core/validation/InvokeArgumentValidator.java
Momo-Not-Emo: > This condition is always false as they point to the same data - we need to get a SootMethod via the view.

The condition is not always true. In the `testInvokeArgumentValidatorFailure` case, the Jimple statement `l1 = virtualinvoke l0.<InvokeArgumentValidator: java.lang.String join(int,java.lang.String)>(1)` is constructed. Here, `invExpr.getArgCount()` evaluates to 1 (indicating one argument passed) while `callee.getParameterTypes().size()` evaluates to 2 (indicating two arguments defined in the method signature). They point to different data.

sootup.tests/src/test/java/sootup/tests/validator/InvokeArgumentValidatorTest.java
Momo-Not-Emo: you are right. This code is copied from an existing validator test. I'll simplify it.

sootup.core/src/main/java/sootup/core/validation/InvokeArgumentValidator.java
swissiety: I see - I was already thinking about the actual method definition and was confused that no retrieving of a method was done.

sootup.core/src/main/java/sootup/core/validation/InvokeArgumentValidator.java
```diff
@@ -22,31 +22,72 @@
...
+                } else {
+                    // check argument type
+                    TypeHierarchy typeHierarchy = view.getTypeHierarchy();
+                    IdentifierFactory identifierFactory = view.getIdentifierFactory();
+                    Iterator<Immediate> iterArgs = args.iterator();
+                    Iterator<Type> iterParameters = parameterTypes.iterator();
+                    while (iterArgs.hasNext() && iterParameters.hasNext()) {
+                        ClassType argType = getClassType(identifierFactory, iterArgs.next().getType());
+                        ClassType paramType = getClassType(identifierFactory, iterParameters.next());
+                        if (argType != paramType && (!typeHierarchy.contains(paramType)
```
swissiety: I am not sure the Typehierarchy can handle primitive type relations / BoxedType compatibility - maybe we have to do that seperately via [type promotion](https://docs.oracle.com/javase/specs/jls/se11/html/jls-5.html)

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/ArchiveBasedAnalysisInputLocation.java
```diff
@@ -99,8 +99,13 @@ public Optional<JavaSootClassSource> getClassSource(@Nonnull ClassType type, @No
...
       FileSystem fs = fileSystemCache.get(path);
       final Path archiveRoot = fs.getPath("/");
+      JavaClassType javaClassType = null;
+      if (type instanceof JavaClassType)
+        javaClassType = (JavaClassType) type;
+      else
+        javaClassType = new JavaClassType(type.getClassName(), type.getPackageName());
       return getClassSourceInternal(
-          (JavaClassType) type, archiveRoot, new AsmJavaClassProvider(view));
+          javaClassType, archiveRoot, new AsmJavaClassProvider(view));
```
swissiety: what happened here?

sootup.java.core/src/main/java/sootup/java/core/JavaIdentifierFactory.java
```diff
@@ -237,6 +237,8 @@ public Collection<PrimitiveType> getAllPrimitiveTypes() {
   @Nonnull
   public JavaClassType getBoxedType(@Nonnull PrimitiveType primitiveType) {
     String name = primitiveType.getName();
+    if (primitiveType instanceof PrimitiveType.IntType)
+      name = "integer";
```
swissiety: lets add a getBoxedName() to the primitive types?

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/ArchiveBasedAnalysisInputLocation.java
Momo-Not-Emo: > what happened here?

It throws exception that an **anonymous** ClassType cannot be cast to JavaClassType:

```java
      ClassType classTypeFieldRefValidator =
                new ClassType() {
                    @Override
                    public boolean isBuiltInClass() {
                        return false;
                    }

                    @Override
                    public String getFullyQualifiedName() {
                        return "jimple.InvokeArgumentValidator";
                    }

                    @Override
                    public String getClassName() {
                        return "InvokeArgumentValidator";
                    }

                    @Override
                    public PackageName getPackageName() {
                        return new PackageName("jimple");
                    }
                };

       ....

        // rt.jar is required since the validator uses typeHierarchy
        DefaultRTJarAnalysisInputLocation defaultRTJarAnalysisInputLocation = new DefaultRTJarAnalysisInputLocation();
        jimpleView = new JimpleView(Arrays.asList(jimpleInputLocation, defaultRTJarAnalysisInputLocation));

        final Optional<SootClass> classSource1 = jimpleView.getClass(classTypeFieldRefValidator);
        assertFalse(classSource1.isPresent());
```

It will search for the `classTypeFieldRefValidator` within both application classes and library classes (rt.jar). Subsequently, it will prompt the casting exception.

In the SootUp class hierarchy, `ClassType` only has one specific type under it, which is `JavaClassType`. I've restricted the search to only look within the application classes. As a result, it won't send an anonymous  `ClassType` object to `ArchiveBasedAnalysisInputLocation` anymore. Should we keep this type check ?

sootup.core/src/main/java/sootup/core/validation/InvokeArgumentValidator.java
Momo-Not-Emo: `Typehierarchy` doesn't consider "int" to be an instance of "Number", that's why I cast the primitive type to its boxed type in "getClassType(...)". 

Actually, in `Typehierarchy.isSubType(...)`, the comment says "subtyping applies to ReferenceTypes only". There is a `PrimitiveHierarchy` class, but its static method `isAncestor(...)` only checks the ancestor-relationship between two primitive types

swissiety: But this does not handle cases like: `int` is used as an argument of a `double` parameter (while double -> int is not allowed implicitly) - we should not fail in these cases

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/ArchiveBasedAnalysisInputLocation.java
swissiety: We should address this issue in whole in another PR - please add a comment that this is a temporary hack

sootup.core/src/main/java/sootup/core/validation/InvokeArgumentValidator.java
Momo-Not-Emo: > But this does not handle cases like: `int` is used as an argument of a `double` parameter (while double -> int is not allowed implicitly) - we should not fail in these cases

I have added implicit conversion checks

sootup.core/src/main/java/sootup/core/validation/InvokeArgumentValidator.java
```diff
@@ -22,31 +22,80 @@
...
+                    while (iterArgs.hasNext() && iterParameters.hasNext()) {
+                        // handle implicit conversion cases. e.g., `int` is used as an argument of a `double` parameter
+                        Type argType = iterArgs.next().getType();
+                        Type paraType = iterParameters.next();
+                        if (argType instanceof PrimitiveType && paraType instanceof PrimitiveType
+                                && PrimitiveType.isImplicitlyConvertibleTo((PrimitiveType) argType, (PrimitiveType) paraType)) {
+                            continue;
+                        }
+                        // other cases
+                        ClassType argClassType = getClassType(identifierFactory, argType);
```
JonasKlauke: After the primitive type check the argument could still have the type: arraytype. You can not assume it is automatically a classtype.

sootup.tests/src/test/resources/validator/jimple/InvokeArgumentValidator.jimple
```diff
@@ -0,0 +1,77 @@
...
+    public void <init>()
+    {
+        InvokeArgumentValidator l0;
+
+        l0 := @this: InvokeArgumentValidator;
+        specialinvoke l0.<java.lang.Object: void <init>()>();
+
+        return;
+    }
+
```
JonasKlauke, line 12: add a testcase for array types. Primitive type array, Classtype array and both cases also as multi-dimensional array

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/ArchiveBasedAnalysisInputLocation.java
```diff
@@ -99,8 +99,15 @@ public Optional<JavaSootClassSource> getClassSource(@Nonnull ClassType type, @No
     try {
       FileSystem fs = fileSystemCache.get(path);
       final Path archiveRoot = fs.getPath("/");
+      JavaClassType javaClassType = null;
+
+      // This is a temporary workaround to prevent a casting exception when passing an anonymous ClassType object
+      if (type instanceof JavaClassType)
```
JonasKlauke: What type do you get here in case of an anoynmous class?

JonasKlauke (changes requested):
Thanks for the help to improve SootUp 👍

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/ArchiveBasedAnalysisInputLocation.java
Momo-Not-Emo: > What type do you get here in case of an anoynmous class?
```
 ClassType classTypeFieldRefValidator =  new ClassType() {...}. // implement the abstract methods
```
It is a ClassType object but not an instance of JavaClassType. I observed that this code has been removed in the current test code.

PR #866, opened 2024-02-22
Adapt method validator and add test cases

Momo-Not-Emo:


sootup.core/src/main/java/sootup/core/validation/MethodValidator.java
```diff
@@ -22,35 +22,50 @@
...
-  public boolean isBasicValidator() {
-    return true;
-  }
+    public static final String staticInitializerName = "<clinit>";
+    @Override
+    public List<ValidationException> validate(@Nonnull Body body, @Nonnull View view) {
+        List<ValidationException> exceptions = new ArrayList<>();
+
+        MethodSignature methodSignature = body.getMethodSignature();
+        SootMethod methodRef = view.getMethod(methodSignature).get();
```
JonasKlauke: The case is not handled that the view does not contain the method

JonasKlauke (changes requested):
Thanks for the help 👍

JonasKlauke (approved):

PR #867, opened 2024-02-22
Adapt IdentityValidator with test cases

Momo-Not-Emo:
The Identity Validator, based on Soot code, now includes added test cases covering all branches, except for cases where "parameterRef" is out-of-index. This omission is intentional, as it prevents the jimple file from being unloaded due to such jimple methods.


sootup.core/src/main/java/sootup/core/validation/IdentityValidator.java
```diff
@@ -33,40 +40,65 @@
...
+     * @return
      */
-    return null;
-  }
+    @Override
+    public List<ValidationException> validate(Body body, View view) {
+        List<ValidationException> exceptions = new ArrayList<>();
+
+        boolean hasThisLocal = false;
+        SootMethod method = view.getMethod(body.getMethodSignature()).get();
```
JonasKlauke: handle the case in which the method can not be found in the given view.

sootup.core/src/main/java/sootup/core/validation/IdentityValidator.java
```diff
@@ -33,40 +40,65 @@
...
+                    if (ref.getIndex() < 0 || ref.getIndex() >= paramCount) {
+                        if (paramCount == 0) {
+                            exceptions.add(new ValidationException(id,
+                                    "This methodRef has no parameters, so no parameter reference is allowed"));
+                        } else {
+                            exceptions.add(new ValidationException(id,
+                                    String.format("Parameter reference index must be between 0 and %d (inclusive)",
+                                            paramCount - 1)));
+                        }
+                        return exceptions;
```
JonasKlauke: if the method also has no this reference, it won't list the exceptions in the results

JonasKlauke (changes requested):
Thanks for the help 👍 
I spotted two improvements. It would be nice if you could include them

sootup.core/src/main/java/sootup/core/validation/IdentityValidator.java
Momo-Not-Emo: > handle the case in which the method can not be found in the given view.

what's the expected behaviour when the method does not present in the given view? Skip this validator or throw an exception?

Momo-Not-Emo:
> Thanks for the help 👍 I spotted two improvements. It would be nice if you could include them

@JonasKlauke I have made changes according to your review. Could you review the changes? Thx~

sootup.core/src/main/java/sootup/core/validation/IdentityValidator.java
swissiety: We should not fail - just abort - we can't say anything about it and in this case we should not complain as we could be in in a partial analysis situation - i.e. SootUp should do the most of the available information

JonasKlauke (approved):
Thanks 👍

JonasKlauke:
Always run the fmt:format plugin before commiting or pushing 😄 I have also created so many fmt commits 😆

PR #871, opened 2024-02-26
Use more precise type

Liyw979:
```@Nonnull private final Immediate[] args; ``` args is already `Immedediate` array.


swissiety (approved):
:+1:

PR #872, opened 2024-02-26
Adapt traps validator from old soot with a test case

Momo-Not-Emo:
It appears challenging to construct a failure test case for the trap validator when striving to ensure the successful loading of Jimple code.


swissiety:
Hi, In the current implementation of `getTraps()` it should be impossible to create such a case as getTraps builds the information from the blocks. Therefore, unfortunately this validator is not needed in SootUp anymore - but lets add it for completeness sake and if someone creates another StmtGraph implementation that works differently/e.g. like Soot. 
We should add a comment then, that implementations based on the MutableBlockStmtGraph should not need this validator.

JonasKlauke (approved):
Thanks 👍

PR #879, opened 2024-02-28
name the Local for the "this" reference: this

swissiety:
+ name "this" this 
+ handle possible `Local` name collisions so that we can distinguish Locals that refer to different Locals/Memory


JonasKlauke (approved):
Thanks

PR #881, opened 2024-02-28
precompute hashCode for JavaClassType and SootClassMemberSignature

Liyw979:
This could look weird.
In my use case, I am using `MethodSignature` as key for caches like `Map<MethodSignature, xxx>` and found out that when doing code analysis,  the cache is frequently read and about 15% of time is spent on `hashCode()`.
Since `MethodSignature` and `JavaClassType` are immutable, the hashCode will not change so I cache them in this PR.


swissiety (approved):
:+1: this seems very reasonable.
I think its not really necessary to do it on demand like its proposed in [effective Java reference on Stackoverflow](https://stackoverflow.com/questions/18473071/how-caching-hashcode-works-in-java-as-suggested-by-joshua-bloch-in-effective-jav) as we use those methods basically all the time as keys.

PR #882, opened 2024-02-28
Make JavaView more functional style

Liyw979:
by using flatMap


JonasKlauke (approved):
Thanks for the help 👍

PR #919, opened 2024-04-11
Add deterministic printing option

stschott:


swissiety (commented):
can we do that less bloated? Maybe act on the option in a separate method

sootup.core/src/main/java/sootup/core/util/printer/JimplePrinter.java
```diff
@@ -191,19 +188,15 @@ public void printTo(SootClass cl, PrintWriter out, LabeledStmtPrinter printer) {
...
-      Iterator<? extends Field> fieldIt;
-      if (options.contains(Option.Deterministic)) {
-        List<? extends Field> fieldList = new ArrayList<>(cl.getFields());
-        fieldList.sort(Comparator.comparing(field -> field.getSignature().toString()));
-        fieldIt = fieldList.iterator();
-      } else {
-        fieldIt = cl.getFields().iterator();
-      }
+      Iterator<? extends SootField> fieldIt =
+          options.contains(Option.Deterministic)
```
swissiety: how about moving the whole construct into it?

swissiety (approved):

swissiety (approved):
:+1:

PR #921, opened 2024-04-13
update implementation for `BackwardsStmtGraph` and add `PostDominanceFinder` utility

shenjunjiekoda:
update the implementation for `BackwardsStmtGraph` in `sootup.core.graph`

before in `DominanceFinder.java`:
```java
List<BasicBlock<?>> preds = new ArrayList<>(block.getPredecessors());
```

after in `DominanceFinder.java`:
```java
List<BasicBlock<?>> preds = new ArrayList<>(blockGraph.predecessors(block));
```

Apply this change , we can create `PostDominaceFinder` as a wrapper of `DominanceFinder`  directly  through use `backwardStmtGraph` as  internal `backingGraph`. To implement this, I  add `successors`/`predecessors` api  in `StmtGraph` and its subclasses  which use block as param instead of dependending on the `BasicBlock` api:  `getPredecessors`/`getSuccessors`.

```java
public class PostDominanceFinder extends DominanceFinder {

  public PostDominanceFinder(StmtGraph<?> blockGraph) {
    super(new BackwardsStmtGraph(blockGraph));
  }
}
```

Thank you for considering this update!

### DIFF
--- a/sootup.core/src/main/java/sootup/core/graph/BackwardsStmtGraph.java
+++ b/sootup.core/src/main/java/sootup/core/graph/BackwardsStmtGraph.java
@@ -23,7 +23,6 @@ package sootup.core.graph;
  */
 import java.util.*;
 import javax.annotation.Nonnull;
-import sootup.core.jimple.basic.Trap;
 import sootup.core.jimple.common.stmt.Stmt;
 
 /** @author Zun Wang */
@@ -49,21 +48,16 @@ public class BackwardsStmtGraph extends ForwardingStmtGraph {
     return Collections.unmodifiableCollection(backingGraph.getNodes());
   }
 
-  @Override
-  public boolean containsNode(@Nonnull Stmt node) {
-    return backingGraph.containsNode(node);
-  }
-
   @Nonnull
   @Override
   public List<Stmt> predecessors(@Nonnull Stmt node) {
-    return successors(node);
+    return backingGraph.successors(node);
   }
 
   @Nonnull
   @Override
   public List<Stmt> successors(@Nonnull Stmt node) {
-    return predecessors(node);
+    return backingGraph.predecessors(node);
   }
 
   @Override
@@ -79,11 +73,5 @@ public class BackwardsStmtGraph extends ForwardingStmtGraph {
   @Override
   public boolean hasEdgeConnecting(@Nonnull Stmt source, @Nonnull Stmt target) {
     return backingGraph.hasEdgeConnecting(target, source);
-  }
-
-  @Nonnull
-  @Override
-  public List<Trap> getTraps() {
-    return backingGraph.getTraps();
   }
 }
--- a/sootup.core/src/main/java/sootup/core/graph/DominanceFinder.java
+++ b/sootup.core/src/main/java/sootup/core/graph/DominanceFinder.java
@@ -40,7 +40,33 @@ public class DominanceFinder {
   private int[] doms;
   private ArrayList<Integer>[] domFrontiers;
 
-  public DominanceFinder(StmtGraph<?> blockGraph) {
+  protected AnalysisDirection direction = AnalysisDirection.FORWARD;
+
+  public enum AnalysisDirection {
+    BACKWARD {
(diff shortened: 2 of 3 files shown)


sootup.core/src/main/java/sootup/core/graph/BackwardsStmtGraph.java
```diff
@@ -51,19 +55,31 @@ public Collection<Stmt> getNodes() {
 
   @Override
   public boolean containsNode(@Nonnull Stmt node) {
-    return backingGraph.containsNode(node);
+    return super.containsNode(node);
   }
```
swissiety: lets remove this method.

sootup.core/src/main/java/sootup/core/graph/BackwardsStmtGraph.java
```diff
@@ -27,15 +27,19 @@
...
 
   @Override
   public Stmt getStartingStmt() {
-    throw new UnsupportedOperationException();
+    List<Stmt> startStmts = getStartingStmts();
+    if (startStmts.isEmpty()) {
+      return null;
+    }
+    return startStmts.get(0);
```
swissiety: these should return the tails of the control flow graph - which could be multiple elements

swissiety (commented):
Thanks for your contribution!
We are trying to introduce only what is really necessary to the API to not run into Soot problems with SootUp - I think adding all those successor/predecessor methods pollutes the API a bit. How about adding a method inside the `Post-` and `DominanceFinder` to initialize the predecessors they way they are needed by the respective *DominanceFinder - i think that would be a simpler & cleaner approach :)

sootup.core/src/main/java/sootup/core/graph/BackwardsStmtGraph.java
shenjunjiekoda: got it

sootup.core/src/main/java/sootup/core/graph/BackwardsStmtGraph.java
shenjunjiekoda: Got it

sootup.core/src/main/java/sootup/core/graph/BackwardsStmtGraph.java
```diff
@@ -35,12 +34,11 @@ public BackwardsStmtGraph(@Nonnull StmtGraph<?> stmtGraph) {
...
-  }
-
-  @Nonnull
-  public List<Stmt> getStartingStmts() {
-    return backingGraph.getTails();
+    List<Stmt> tails = backingGraph.getTails();
+    if (tails.isEmpty()) {
+      return null;
+    }
+    return tails.get(0);
```
swissiety: I think we should keep it with the UnsupportedException instead of arbitrarily choosing one Tail.

sootup.core/src/main/java/sootup/core/graph/DominanceFinder.java
```diff
@@ -76,7 +102,8 @@ public DominanceFinder(StmtGraph<?> blockGraph) {
           continue;
         }
         int blockIdx = blockToIdx.get(block);
-        List<BasicBlock<?>> preds = new ArrayList<>(block.getPredecessors());
+        List<?> predecessors = block.getPredecessors();
```
swissiety: is predecessors used somewhere?

swissiety (commented):
looks good :+1:

sootup.core/src/main/java/sootup/core/graph/BackwardsStmtGraph.java
shenjunjiekoda: Got it

sootup.core/src/main/java/sootup/core/graph/DominanceFinder.java
shenjunjiekoda: Sorry, it's my mistake, I'll remove that

swissiety (approved):
good job!

PR #927, opened 2024-04-18
streamify TypeHierarchy

swissiety:
- [x] return Streams instead of collections - user can decide to collect without SootUp copying it in between  
- [x] steamify visitSubGraph as well (not colelction.stream() internally.. )
- [x] adapt testcases


swissiety:
lgtm

JonasKlauke (approved):

PR #929, opened 2024-04-22
Types validator changes

akshitad11:


sootup.core/src/main/java/sootup/core/validation/TypesValidator.java
```diff
@@ -34,18 +39,48 @@ public class TypesValidator implements BodyValidator {
...
-     * ") local type not allowed in final code: " + t + " local: " + l)); } }
-     */
-    return null;
+    List<ValidationException> validationExceptions = new ArrayList<>();
+
+    Optional<? extends SootMethod> sootMethodOpt = view.getMethod(body.getMethodSignature());
+    if (!sootMethodOpt.isPresent()) {
+      throw new IllegalStateException(
+          "We should find the given method to the given Body in the View. wrong View or Method?");
+    }
```
swissiety: lets fail silently here - work with all the information we have available.

sootup.core/src/main/java/sootup/core/validation/TypesValidator.java
```diff
@@ -34,18 +39,48 @@ public class TypesValidator implements BodyValidator {
...
+    SootMethod method = sootMethodOpt.get();
+
+    if (!(method.getReturnType() instanceof PrimitiveType
+            || method.getReturnType() instanceof VoidType
+            || method.getReturnType() instanceof ReferenceType)
+        || method.getReturnType() instanceof NullType) {
```
swissiety: nulltype should not be allowed (as a method definition can not have the null reference as definition - it is returned for a referencetype - is a different thing)

sootup.core/src/main/java/sootup/core/validation/TypesValidator.java
```diff
@@ -34,18 +39,48 @@ public class TypesValidator implements BodyValidator {
...
+          new ValidationException(
+              method,
+              "Return type not allowed in final code: " + method.getReturnType(),
+              "return type not allowed in final code:"
+                  + method.getReturnType()
+                  + "\n methodRef: "
+                  + method));
+    }
+    for (Type t : method.getParameterTypes()) {
+      if (t instanceof VoidType || !(t instanceof PrimitiveType || t instanceof ReferenceType)) {
```
swissiety: t instanceof VoidType seems redundant to the  conditions on the right

kadirayk (approved):

PR #936, opened 2024-05-17
Rework the call graph algorithm to annote call edges with the corresponding invoke statement

### LINKED ISSUE #483: Code line numbers in call-graphs.
Do we need code line numbers in our generated call graphs?

### LINKED ISSUE #902: Unexpected Clinit calls in the call graph
Hi, I used SootUp to analyze some projects and found false positives which affect my analysis results. After, investigation, I found the following minimized code example:

```java
package org.example;
public class C {
    private static final String CName = C.class.getName();
    public void foo(){}
}
```
SootUp reports an edge from `<org.example.C: void foo()>` to `<org.example.C: void <clinit>()>`, which is incorrect. It seems that this bug is related to the initialization of <clinit> method.

#### SootUp configuration
```java
AnalysisInputLocation<JavaSootClass> javaBaseInputLocation = new JavaClassPathAnalysisInputLocation("Path/to/javaBase", SourceType.Library);
AnalysisInputLocation<JavaSootClass> classInput = new JavaClassPathAnalysisInputLocation("Path/to/classDir", SourceType.Application);
JavaProject project = JavaProject.builder(new JavaLanguage(8))
          .addInputLocation(classInput)
          .addInputLocation(javaBaseInputLocation)
          .build();
JavaView view = project.createView();
CallGraph cg;
String EntrySignature="<org.example.C: void foo()>"
List<MethodSignature> entryMethods = new ArrayList<>();
for (JavaSootClass klass : classes) {
    for (JavaSootMethod method : klass.getMethods()) {
        if (method.getSignature().toString().equals(EntrySignature)) {
            entryMethods.add(method.getSignature());
        }
    }
}
CallGraphAlgorithm cha = new ClassHierarchyAnalysisAlgorithm(Constants.view);
CallGraph cg = cha.initialize(entryMethods);
```

SootUp version: 1.1.2

### LINKED ISSUE #903: An inconsistency between call graphs
I used SootUp to construct call graphs for my project, and found a bug in CHA algorithms.

A.java
```java
package org.sslab;
import java.io.Closeable;
public interface A extends Closeable {
    @Override
    default void close(){
        close();
    }
}
```
B.java
```java
package org.sslab;
public class B implements A {
    public static void main(String[] args) {
        try (B b = new B()) {
        }
    }
}
```

In the above code examples, RTA call graph includes an edge from `B.main` to `A.close` that is reasonable, but CHA does not have. I think CHA should provide a more sound analysis results.

This edge seems related to callback process of SootUp as no type hierarchy and new expressions here to guide the above two algorithms.

SootUp version: 1.1.2
#### Configuration
```java
AnalysisInputLocation<JavaSootClass> javaBaseInputLocation = new JavaClassPathAnalysisInputLocation("Path/to/javaBase", SourceType.Library);
AnalysisInputLocation<JavaSootClass> classInput = new JavaClassPathAnalysisInputLocation("Path/to/classDir", SourceType.Application);
JavaProject project = JavaProject.builder(new JavaLanguage(8))
          .addInputLocation(classInput)
          .addInputLocation(javaBaseInputLocation)
          .build();
JavaView view = project.createView();

String EntrySignature="<org.sslab.B: void main(java.lang.String[])>"
List<MethodSignature> entryMethods = new ArrayList<>();
for (JavaSootClass klass : classes) {
    for (JavaSootMethod method : klass.getMethods()) {
        if (method.isMain() && method.getSignature().toString().equals(EntrySignature)) {
            entryMethods.add(method.getSignature());
        }
    }
}

CallGraphAlgorithm cha = new ClassHierarchyAnalysisAlgorithm(Constants.view);
CallGraph cg1 = cha.initialize(entryMethods);

CallGraphAlgorithm rta = new RapidTypeAnalysisAlgorithm(view);
CallGraph cg2 = rta.initialize(entryMethods);
```

JonasKlauke:
1. introduces Invokable Statements on edges of the call graph to know the target of a specific invokable statement. closes #483 
2. moves clinit invokes from the constructor invoke to corresponding new statements before the call, to support clinit calls caused by arrays or multiarrys
3. call graphs will save a list entry methods
4. clinit calls of given start methods are added to the entry methods of a call graphs
5. the edge between a starting method and their corresponding clinit method was removed fixes #902 
6. added test for bug report #903, fixes #903


JonasKlauke:
entry methods wont contain anymore an edge to the clinit method.
clinit methods will be added as new entry points to the call graph.
fixes #902

sootup.analysis/src/main/java/sootup/analysis/interprocedural/icfg/CGEdgeUtil.java
```diff
@@ -59,12 +60,11 @@ public static Set<Pair<MethodSignature, CalleeMethodSignature>> getCallEdges(
       SootMethod method = view.getMethod(caller).orElse(null);
       if (method != null && method.hasBody()) {
         for (Stmt s : method.getBody().getStmtGraph().getNodes()) {
-          if (s.containsInvokeExpr()) {
+          if (s instanceof InvokableStmt && ((InvokableStmt) s).containsInvokeExpr()) {
```
swissiety: its removes clutter from Stmt which has a (sometimes) exception throwing method (getInvokeExpr()). But from a user perspective it seems weird / maybe not obvious why we check twice for the "same" thing (which it isn't but I'm not sure this is obvious). Maybe we should keep Stmt.containsInvokeExpr().

sootup.analysis/src/test/java/sootup/analysis/interprocedural/ifds/IFDSTaintAnalysisProblem.java
```diff
@@ -74,18 +74,18 @@ public FlowFunction<Value> getNormalFlowFunction(Stmt curr, Stmt succ) {
...
       public FlowFunction<Value> getReturnFlowFunction(
           Stmt callSite, SootMethod calleeMethod, Stmt exitStmt, Stmt returnSite) {
-        return getReturnFlow(callSite, calleeMethod, exitStmt, returnSite);
+        return getReturnFlow((InvokableStmt) callSite, calleeMethod, exitStmt, returnSite);
       }
 
       @Override
       public FlowFunction<Value> getCallToReturnFlowFunction(Stmt callSite, Stmt returnSite) {
-        return getCallToReturnFlow(callSite, returnSite);
+        return getCallToReturnFlow((InvokableStmt) callSite, returnSite);
```
swissiety: can we adapt heros to have different generic types for callSite Stmts and the other Stmts?

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -92,41 +94,32 @@ final CallGraph constructCompleteCallGraph(View view, List<MethodSignature> entr
...
-          if (!cg.containsCall(methodSignature, staticInitSig)) {
-            cg.addCall(methodSignature, staticInitSig);
-            workList.push(staticInitSig);
-          }
-        });
+  protected List<MethodSignature> getClinitFromEntryPoints(List<MethodSignature> entryPoints) {
+    return entryPoints.stream()
+        .map(
+            methodSignature ->
+                getSignatureOfImplementedStaticInitializer(methodSignature.getDeclClassType()))
```
swissiety, line 115:
```suggestion
     methodSignature.getDeclClassType().getStaticInitializer())
```
can we do this or do we have to check if the clinit exists? 
If no, maybe refactor/simplify the method name to "getStaticInitializer" ?

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -200,103 +179,169 @@ final void processWorkList(
...
-   * @param view it contains the class data
    * @param sourceMethod the inspected source method
-   * @return a stream containing all method signatures of targets of implicit calls.
+   * @param cg clinit calls will be added to the call graph
+   * @param workList found clinit methods will be added to the work list
    */
-  @Nonnull
-  protected Stream<MethodSignature> resolveAllStaticInitializerCallsFromSourceMethod(
-      View view, SootMethod sourceMethod) {
+  protected void resolveAllStaticInitializerCallsFromSourceMethod(
```
swissiety: long method name: how about resolveStaticInitializerCalls?

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -200,103 +179,169 @@ final void processWorkList(
...
+   * clinit methods of the targetClass and Superclasses will be added to the call graph. If new
+   * target methods will be found, the worklist will be updated.
+   *
+   * @param sourceSig the source method causing the static initilzer call
+   * @param targetClass the class that is statically initialized
+   * @param invokableStmt the statement causing the call
+   * @param cg the call graph that will contain the found calls
+   * @param workList the work list that will be updated with new target methods
+   */
+  private void addStaticInitializerCallsToCallGraph(
```
swissiety: long name: addStaticInitializerCalls()?

sootup.core/src/main/java/sootup/core/jimple/common/stmt/InvokableStmt.java
```diff
@@ -0,0 +1,55 @@
...
+   * @return true if the statement invokes something, and false if not.
+   */
+  boolean doesInvoke();
+
+  /**
+   * Checks if the invokable statement contains a invoke expression that defines the invoke.
+   *
+   * @return true if the statement contains an invoke expression, false if not
+   */
+  boolean containsInvokeExpr();
```
swissiety: where is the difference?

swissiety: from the implementaion doesInvoke is for static initializer handling as well? Maybe lets rename it to sth like  canInvokeStaticInitializer?

sootup.core/src/main/java/sootup/core/types/ArrayType.java
```diff
@@ -99,4 +99,18 @@ public Type getElementType() {
   public void accept(@Nonnull TypeVisitor v) {
     v.caseArrayType();
   }
+
+  public boolean isArrayTypeOfPrimitives() {
+    return isArrayTypeOfPrimitives(baseType);
+  }
+
+  private boolean isArrayTypeOfPrimitives(Type type) {
```
swissiety: lets make it public static?

swissiety (dismissed):
I'm not sure about introducing all these TestUtil classes - Currently it seems/feels like it is creating a lot of additional clutter which does make understanding tests more complicated.. maybe we can reiterate that and have a look how other projects solve it, especially as of the current implementation we can't share those classes across the submodules and would have code duplicates.

sootup.analysis/src/main/java/sootup/analysis/interprocedural/icfg/CGEdgeUtil.java
JonasKlauke: Not all Invokable stmt contain a Invoke expr. These are not the same. Even doesInvoke is not the same since "int a = 5;" would be also an InvokableStmt which does not invoke anything.
And I thought we wanted to go away from the stmt thingy that throws exceptions for not fitting stmts. On the other side you could say it could be weird for the developer that all other stmts throw an exception?

sootup.analysis/src/test/java/sootup/analysis/interprocedural/ifds/IFDSTaintAnalysisProblem.java
JonasKlauke: I did not touch the analysis stuff. I made it only work for the current state. I will add an issue for the complete adaption. Or should this PR also contain the adaption to analysis?

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
JonasKlauke: If the class wont have any static fields which are set, it wont have a clinit method

sootup.core/src/main/java/sootup/core/jimple/common/stmt/InvokableStmt.java
JonasKlauke: yeah currently it is only for clinit calls caused by the new statement or static fields. yeah we can rename it to static initializer, but this will say that normal virtual invokes would be false

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
JonasKlauke: so no we need the check if the method exists, since getSTaticInitializer just returns a signature without checking the existance

sootup.core/src/main/java/sootup/core/jimple/common/stmt/InvokableStmt.java
JonasKlauke: this is applied

sootup.analysis/src/test/java/sootup/analysis/interprocedural/ifds/IFDSTaintAnalysisProblem.java
JonasKlauke: I created #937

sootup.core/src/main/java/sootup/core/jimple/common/stmt/InvokableStmt.java
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * @return true if the statement invokes something, and false if not.
+   */
+  boolean doesInvoke();
+
+  /** Checks if the invokable statement contains a invoke expression that defines the invoke.
+   *
+   * @return true if the statement contains an invoke expression, false if not
+   */
+  boolean containsInvokeExpr();
```
kadirayk, line 48: If it contains an InvokeExpr isn't it already an InvokeStmt?

JonasKlauke: Nope it is an invokable stmt. Assignments can contain invoke expressions so a = 3 would be also an invokableStmt

kadirayk (approved):

PR #943, opened 2024-06-05
Fix Type.getValueBitSize in case of a CharType

marcus-h:
According to jls-4.2 [1], the value bit size of a char is 16 bit. The old code
returned 32 bit.

Note that the added TypeTest is not executed by default, when running
"mvn test". Since other tests are also affected, I assume this is the
intended behavior.

[1] https://docs.oracle.com/javase/specs/jls/se22/html/jls-4.html#jls-4.2


swissiety (approved):
:+1: nice finding

marcus-h:
I just did a forced push to my branch in order to pacify the style checker.

PR #945, opened 2024-06-06
Avoid a potential CCE in Position.equals

marcus-h:
The old code in sootup.core.model.Position.equals always casts the passed argument to a FullPosition. However, in theory, it is possible that there are multiple subclasses of Position. Hence, the cast could potentially result in a CCE. In order to avoid this, just cast the passed argument to Position itself, which is always safe (due to the previous checks).

### DIFF
--- a/sootup.core/src/main/java/sootup/core/model/Position.java
+++ b/sootup.core/src/main/java/sootup/core/model/Position.java
@@ -56,7 +56,7 @@ public abstract class Position implements Comparable<Position> {
     if (o == null || getClass() != o.getClass()) {
       return false;
     }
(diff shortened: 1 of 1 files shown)


sootup.core/src/main/java/sootup/core/model/Position.java
```diff
@@ -56,7 +56,7 @@ public boolean equals(Object o) {
     if (o == null || getClass() != o.getClass()) {
```
JonasKlauke:
```suggestion
    if (!(o instanceof Position)) {
```

JonasKlauke (approved):

PR #956, opened 2024-06-18
Fix typo in `jimple.md`

jyoo980:
I think "traip" really means "trap" in the context of "traip chain"


swissiety:
Thx for your contribution! I already did a rework of the documentation site in the currenlty open PR #955 and fixed it there.

PR #959, opened 2024-06-21
add TypeHierarchy doc

swissiety:


JonasKlauke (approved):

sootup.core/src/main/java/sootup/core/typehierarchy/TypeHierarchy.java
```diff
@@ -75,6 +75,7 @@ public interface TypeHierarchy {
    * For an interface type, this does the same as {@link #implementersOf(ClassType)}. For a class
    * type, this does the same as {@link #subclassesOf(ClassType)}.
    */
+  // TODO: [ms] check! not sure this method makes sense in the interface..
```
JonasKlauke, line 78: Interfaces can extend from each other, so a subtype of an interface are interfaces.
It could be also that this methods also returns classes that implement the given interface

swissiety: * if we need it semantically - it seemed like a method that was extracted from the implementation class

PR #972, opened 2024-07-05
Fix #960 ushr get type

### LINKED ISSUE #960: [Bug]: Potential Issue in getType Method of JUshrExpr.java
### What happened?

Dear SootUp Developers,

I have identified a potential issue in the `getType` method of `JUshrExpr` located at [JUshrExpr.java: L56](https://github.com/soot-oss/SootUp/blob/develop/sootup.core/src/main/java/sootup/core/jimple/common/expr/JUshrExpr.java#L56).

In [the original Soot code](https://github.com/soot-oss/soot/blob/develop/src/main/java/soot/jimple/internal/JUshrExpr.java#L60), the `getType` method is implemented as follows:

```java
@Override
public Type getType() {
  if (isIntLikeType(op2Box.getValue().getType())) {
    final Type t1 = op1Box.getValue().getType();
    if (isIntLikeType(t1)) {
      return IntType.v();
    }
    final LongType tyLong = LongType.v();
    if (tyLong.equals(t1)) {
      return LongType.v();
    }
  }
  return UnknownType.v();
}
```

However, in the refactored version in SootUp, the code is:

```java
@Nonnull
@Override
public Type getType() {
  Value op1 = getOp1();
  Value op2 = getOp2();

  if (Type.isIntLikeType(op2.getType())) {
    return UnknownType.getInstance();
  }

  if (Type.isIntLikeType(op1.getType())) {
    return PrimitiveType.getInt();
  }
  if (op1.getType().equals(PrimitiveType.getLong())) {
    return PrimitiveType.getLong();
  }

  return UnknownType.getInstance();
}
```

I believe there is a logical error in the condition: 

```java
if (Type.isIntLikeType(op2.getType())) {
  return UnknownType.getInstance();
}
```

It should be:

```java
if (!Type.isIntLikeType(op2.getType())) {
  return UnknownType.getInstance();
}
```

This is consistent with other similar `getType` method implementations, such as the one found in [JShrExpr.java: L57](https://github.com/soot-oss/SootUp/blob/develop/sootup.core/src/main/java/sootup/core/jimple/common/expr/JShrExpr.java#L57).

### Version

Latest develop branch

### Relevant log output

_No response_

swissiety:
closes #960

### DIFF
--- a/sootup.core/src/main/java/sootup/core/jimple/common/expr/AbstractFloatBinopExpr.java
+++ b/sootup.core/src/main/java/sootup/core/jimple/common/expr/AbstractFloatBinopExpr.java
@@ -44,14 +44,19 @@ public abstract class AbstractFloatBinopExpr extends AbstractBinopExpr {
 
     if (Type.isIntLikeType(op1t) && Type.isIntLikeType(op2t)) {
       return PrimitiveType.getInt();
-    } else if (op1t.equals(PrimitiveType.getLong()) || op2t.equals(PrimitiveType.getLong())) {
-      return PrimitiveType.getLong();
-    } else if (op1t.equals(PrimitiveType.getDouble()) || op2t.equals(PrimitiveType.getDouble())) {
-      return PrimitiveType.getDouble();
-    } else if (op1t.equals(PrimitiveType.getFloat()) || op2t.equals(PrimitiveType.getFloat())) {
-      return PrimitiveType.getFloat();
-    } else {
-      return UnknownType.getInstance();
     }
+    final PrimitiveType.LongType longType = PrimitiveType.getLong();
+    if (op1t == longType || op2t == longType) {
+      return longType;
+    }
+    final PrimitiveType.DoubleType aDouble = PrimitiveType.getDouble();
+    if (op1t == aDouble || op2t == aDouble) {
+      return aDouble;
+    }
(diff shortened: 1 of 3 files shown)


sootup.core/src/main/java/sootup/core/jimple/common/expr/AbstractFloatBinopExpr.java
```diff
@@ -44,14 +44,19 @@ public Type getType() {
...
+      return longType;
+    }
+    final PrimitiveType.DoubleType aDouble = PrimitiveType.getDouble();
+    if (op1t == aDouble || op2t == aDouble) {
+      return aDouble;
+    }
+    final PrimitiveType.FloatType aFloat = PrimitiveType.getFloat();
+    if (op1t == aFloat || op2t == aFloat) {
+      return aFloat;
+    }
```
stschott, line 59: We should name the other types `doubleType` and `floatType` as well, just to stay consistent.
Also, are you sure that `==` is sufficient here? The types are not implemented as enums.

stschott (approved):

sootup.core/src/main/java/sootup/core/jimple/common/expr/AbstractFloatBinopExpr.java
swissiety: but as singletons - on the other hand in theory someone could subclass them..

PR #974, opened 2024-07-08
enable to chain the acceptors with the given visitor

swissiety:
1)
this would enable to use:

`stmt.accept(visitor).getResult();`

instead of:

```java
stmt.accept(visitor);
visitor.getResult();
```

2) remove mandatory result field + getters + setter


sootup.core/src/main/java/sootup/core/types/ArrayType.java
```diff
@@ -96,7 +96,8 @@ public Type getElementType() {
   }
 
   @Override
-  public void accept(@Nonnull TypeVisitor v) {
+  public TypeVisitor accept(@Nonnull TypeVisitor v) {
```
swissiety: oopsie. // TODO

kadirayk (approved):

PR #993, opened 2024-07-23
Create test_jars workflow

palaniappan1:


swissiety:
Maybe it makes sense to keep a history of artifacts so that we can exclude these artifacts from the random selection?

kadirayk (approved):

PR #995, opened 2024-07-24
add null check in removeAllEdgesOutOf method

### LINKED ISSUE #994: [Bug]: Potential Missing Nulll Check in Soot but not in SootUp
## Description:
I have identified an instance where a null check present in the original Soot project are missing in SootUp. These methods have a direct correspondence between Soot and SootUp. Missing checks could cause issues in the future, as they should exhibit similar behavior.

### Method `removeAllEdgesOutOf`
- Mapped Method in SootUp:
  - Class: `qilin.core.builder.callgraph.OnFlyCallGraph`
  - [Link to Code](https://github.com/soot-oss/SootUp/blob/develop/sootup.qilin/src/main/java/qilin/core/builder/callgraph/OnFlyCallGraph.java#L109)
    ```java
    for (QueueReader<Edge> edgeRdr = listener(); edgeRdr.hasNext(); ) {
      Edge e = edgeRdr.next();
      if (e.srcUnit() == u) {
        e.remove();
        removeEdge(e, false);
        edgesToRemove.add(e);
        hasRemoved = true;
      }
    }
    ```
- Corresponding Method in Soot:
  - Class: `soot.jimple.toolkits.callgraph.CallGraph`
  - [Link to Code](https://github.com/soot-oss/soot/blob/develop/src/main/java/soot/jimple/toolkits/callgraph/CallGraph.java#L103)
    ```java
    for (QueueReader<Edge> edgeRdr = listener(); edgeRdr.hasNext();) {
      Edge e = edgeRdr.next();
      if (e != null && e.srcUnit() == u) {
        e.remove();
        removeEdge(e, false);
        edgesToRemove.add(e);
        hasRemoved = true;
      }
    }
    ```
In the original Soot implementation, there is a null check for `e` before calling `e.srcUnit() == u`. However, this null check is omitted in the SootUp implementation. I also verified that `srcUnit()` doesn't have any annotations indicating that it cannot return null, which means this omission could lead to a potential NullPointerException (NPE).

### Version

Latest develop branch

### Relevant log output

_No response_

Momo-Not-Emo:
fix #994

### DIFF
--- a/sootup.qilin/src/main/java/qilin/core/builder/callgraph/OnFlyCallGraph.java
+++ b/sootup.qilin/src/main/java/qilin/core/builder/callgraph/OnFlyCallGraph.java
@@ -108,7 +108,7 @@ public class OnFlyCallGraph implements MutableCallGraph, Iterable<Edge> {
     Set<Edge> edgesToRemove = new HashSet<>();
     for (QueueReader<Edge> edgeRdr = listener(); edgeRdr.hasNext(); ) {
       Edge e = edgeRdr.next();
-      if (e.srcUnit() == u) {
+      if (e != null && e.srcUnit() == u) {
         e.remove();
         removeEdge(e, false);
         edgesToRemove.add(e);


kadirayk (approved):
Thanks a lot for the improvement!

PR #997, opened 2024-07-28
issue-982 : Assign TopType to ObjectType at end of TypeAssigner

### LINKED ISSUE #982: [Bug]: TypeAssigner fails if LocalSplitter is not applied beforehand
### What happened?

When using the TypeAssigner without the LocalSplitter there are some cases where the TypeAssigner assignes `BottomType` or `TopType`. Make the TypeAssigner not assign any type in such cases to not break the Jimple. 

### Version

Latest develop branch

### Relevant log output

_No response_

sahilagichani14:
assign TopType to ObjectType at the end of TypeAssigner so LocalNameStandardizer doesn't give exception after.


shared-test-resources/TypeResolverTestSuite/Misc/Misc.java
```diff
@@ -70,6 +72,35 @@ public static void impossibleTyping() {
...
 
+  public static void testTaAndLnsWithoutLS() {
+    int a = 10, c;
+    try {
+      a = 8 / 0;
+      int b = a + 2;
+      c = a;
+    } catch (ArithmeticException ex) {
+    }
+    System.out.println(a);
```
JonasKlauke: remove println from the testcases. It unnecessary blows up the tests

shared-test-resources/TypeResolverTestSuite/Misc/io/jsonwebtoken/impl/lang/JwtDateConverter.class
stschott, line 1: Please isolate the issue and provide a source/class file pair instead of a class from a different library.

stschott (changes requested):

shared-test-resources/TypeResolverTestSuite/Misc/io/jsonwebtoken/impl/lang/JwtDateConverter.class
sahilagichani14: I have added that test case as well, I will remove this class from a diff lib.

stschott (approved):

PR #1006, opened 2024-08-06
Remove dead link to wiki in README

stschott:


swissiety (commented):
lets encourage / state we are happy to accept contributions - to prevent frustrations its best to open an issue/discussion what you plan to do, so that possible problems or unnecessary work can be eradicated as early as possible? sth like that..

PR #1022, opened 2024-08-13
modified CopyPropagator to function as expected

### LINKED ISSUE #1021: [Bug]: CopyPropagator not Propagating Constant to sout, binary expressions, if condition
### What happened?

CopyPropagator not Propagating Constant to sout, binary expressions, if condition

### Version

Latest develop branch

### Relevant log output

_No response_

sahilagichani14:


swissiety (approved):
:+1: nice finding!

sootup.java.core/src/main/java/sootup/java/core/interceptors/CopyPropagator.java
```diff
@@ -86,25 +86,27 @@ else if (rhs instanceof JCastExpr && rhs.getType() instanceof ReferenceType) {
...
+          newStmt = replaceUse(stmtGraph, newStmt, use, rhs);
         }
       }
     }
   }
 
-  private void replaceUse(
+  private Stmt replaceUse(
       @Nonnull MutableStmtGraph graph, @Nonnull Stmt stmt, @Nonnull Value use, @Nonnull Value rhs) {
     if (!use.equivTo(rhs)) { // TODO: ms: check if rhs!=use would be enough
```
swissiety: can you check that TODO as well?

sahilagichani14:
https://arxiv.org/pdf/2207.03894, 
Definition 3[Copy Propagation]: The use of a variable y in the statement z=x+y occurring at a point
p can be replaced by a variable w if every path from the entry node to point p contains the same
definition y=w, for the variable y, and after the definition prior to reaching p, there is no redefinition to
the variable y and **no redefinition to the variable w**.

sootup.java.core/src/main/java/sootup/java/core/interceptors/CopyPropagator.java
```diff
@@ -86,25 +87,68 @@ else if (rhs instanceof JCastExpr && rhs.getType() instanceof ReferenceType) {
...
+            newStmt = replaceUse(stmtGraph, newStmt, use, NullConstant.getInstance());
           }
         }
         // if rhs is a local, then replace use, if it is possible
         else if (rhs instanceof Local && !rhs.equivTo(use)) {
-          replaceUse(stmtGraph, stmt, use, rhs);
+          Local m = (Local) rhs;
+          if (use != m) {
+            Integer defCount = m.getDefs(stmtGraph.getStmts()).size();
+            if (defCount == null || defCount == 0) {
```
swissiety: defCount cannot be null

sootup.java.core/src/main/java/sootup/java/core/interceptors/CopyPropagator.java
```diff
@@ -86,25 +87,68 @@ else if (rhs instanceof JCastExpr && rhs.getType() instanceof ReferenceType) {
...
         }
       }
     }
   }
 
-  private void replaceUse(
+  private Stmt replaceUse(
       @Nonnull MutableStmtGraph graph, @Nonnull Stmt stmt, @Nonnull Value use, @Nonnull Value rhs) {
-    if (!use.equivTo(rhs)) { // TODO: ms: check if rhs!=use would be enough
+    if (!use.equivTo(rhs) || rhs != use) {
```
swissiety: the comment was meant to replace the current conditionExpr - not to add it as well. (if `! equivto()` is false then rhs should be != use as well)

sahilagichani14: I tested the test cases that I had and seems to work fine with rhs!=use.

sootup.java.core/src/main/java/sootup/java/core/interceptors/CopyPropagator.java
```diff
@@ -86,25 +87,68 @@ else if (rhs instanceof JCastExpr && rhs.getType() instanceof ReferenceType) {
...
           }
         }
         // if rhs is a local, then replace use, if it is possible
         else if (rhs instanceof Local && !rhs.equivTo(use)) {
-          replaceUse(stmtGraph, stmt, use, rhs);
+          Local m = (Local) rhs;
+          if (use != m) {
+            Integer defCount = m.getDefs(stmtGraph.getStmts()).size();
+            if (defCount == 0) {
+              throw new RuntimeException("Variable " + m + " used without definition!");
```
swissiety: IllegalState

swissiety (approved):
please adapt the exception type and we are good to go

sootup.java.core/src/main/java/sootup/java/core/interceptors/CopyPropagator.java
swissiety:
```suggestion
              throw new IllegalStateException("Local `" + m + "' is used without a definition!");
```

PR #1025, opened 2024-08-15
Implement call graph pruning feature

bingtimren:
Generating large call graph can be slow and consumes large memory.

This feature allows pruning the call graph with a bound function, so to generate an incomplete call graph according to the analysis algorithm user's specific needs - with potentially less memory & CPU usage.


sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -232,7 +235,7 @@ protected void resolveAllCallsFromSourceMethod(
         .map(Stmt::asInvokableStmt)
         .forEach(
             stmt ->
-                resolveCall(sourceMethod, stmt)
+                 (boundFunction == null || Boolean.TRUE.equals(boundFunction.apply(sourceMethod, stmt)) ? resolveCall(sourceMethod, stmt) : Stream.<MethodSignature> empty())
```
swissiety: 1) instead of null lets make use of a method that just returns always true
2) lets use BiPredicate as type to reduce unnecessary AutoBoxing and simplify the condition check  to just the method call

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -586,4 +589,25 @@ public static Optional<SootMethod> findConcreteMethod(
...
+   * the source method to the invocation target shall be added into the call graph.
+   * 
+   * This allows building an pruned & potentially incomplete call graph, with lower memory / CPU
+   * cost.
+   * 
+   * @param boundFunction */
+  public void setBoundFunction(BiFunction<SootMethod, InvokableStmt, Boolean> boundFunction)
+  {
+      this.boundFunction = boundFunction;
+  }
```
swissiety: lets set the function in the constructor only

swissiety (commented):
thanks for that contribution! I will merge it with just some small adjustments.

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
bingtimren: will make adjustment as suggested

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
bingtimren: hmm.... but this would force all sub-classes implement two constructors - one with a bound function, one without (so to use the always-true function)..... instead of having this feature automatically

let's keep it this way, or change setter to a "with" method (fluent API)?

bingtimren: if you are concerned bound function is set after constructCompleteCallGraph() is called, I can add a checker and throw an exception if user do that. thought?

swissiety: lets implement it as a method directly without BiPredicate etc. - that way someone can override it the same way its done with a linkedhashmap to transform it to a LRUcache via overriding removeoldestentry() - e.g. see https://ashcode.medium.com/lru-cache-using-linkedhashmap-java-85e78ee1b976

bingtimren: I would still recommend keeping it this way. This is more a setting of the analyser than an extension to it. In general I would try to avoid overriding a non-abstract method, it's kind of breaking encapsulation. In general composition is preferred than inheritance, see https://en.wikipedia.org/wiki/Composition_over_inheritance

JonasKlauke: I think if you configure it in the constructor, the developer might be more aware of the feature.

sootup.callgraph/src/test/java/sootup/callgraph/RapidTypeAnalysisPruningAlgorithmTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.Collections;
+import org.junit.jupiter.api.Test;
+import sootup.core.model.SootClass;
+import sootup.core.model.SootMethod;
+import sootup.core.signatures.MethodSignature;
+import sootup.java.core.views.JavaView;
+
+public class RapidTypeAnalysisPruningAlgorithmTest
+    extends CallGraphTestBase<RapidTypeAnalysisPruningAlgorithm> {
```
JonasKlauke: Solution without extends.
Maybe use Util Class for views and so on

JonasKlauke (approved):

PR #1043, opened 2024-08-27
runtime & memory usage to body interceptors

### LINKED ISSUE #1035: [Feature]: Get Performance of BodyInterceptors (Runtime and Memory Consumption)
### Missing Feature

As a user, I want to know how much time and memory each body interceptor uses so I can choose which ones to apply in the view for better performance.

sahilagichani14:


sootup.core/src/main/java/sootup/core/transform/BodyInterceptorMetric.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ * #L%
+ */
+
+public class BodyInterceptorMetric {
+
+  private Long runtime;
```
palaniappan1: It would be better to use long (primitive type) instead of Long (Object type), I don't see the usecase of using Object type here.

sootup.core/src/main/java/sootup/core/transform/RunTimeBodyInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+    long usedMemoryBefore = runtime.totalMemory() - runtime.freeMemory();
+
+    bodyInterceptor.interceptBody(builder, view);
+
+    long endTime = System.currentTimeMillis(); // End time
+    long duration = endTime - startTime;
+    long usedMemoryAfter = runtime.totalMemory() - runtime.freeMemory();
+    long memoryUsed = (usedMemoryAfter - usedMemoryBefore) / MB;
+
+    biMetric.setRuntime(biMetricMap.get(bodyInterceptor).getRuntime() + duration);
```
sahilagichani14: Maybe in the future for Metrics objects, we might need to modify Long objects and here as well I'm setting object values for runtime and memory usage hence I used Long in BodyInterceptorMetric, and inside the runtime interceptBody calculation of that is fixed and I'm just reading the value hence I used primitive long here. @palaniappan1

palaniappan1:
FYI, java.util.* has almost 70-80 classes & interfaces included. You just used 2 or 3 classes, then why do we need that whole package imported. I cleaned it up, so no worries :)

palaniappan1 (approved):

PR #1045, opened 2024-08-29
Added new inputlocation for downloading jars from maven

palaniappan1:


sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/DownloadJarAnalysisInputLocation.java
```diff
@@ -0,0 +1,51 @@
...
+        int bytesRead;
+        while ((bytesRead = inputStream.read(buffer, 0, BUFFER_SIZE)) != -1) {
+          outputStream.write(buffer, 0, bytesRead);
+        }
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+
+    return file.exists() ? file.toPath() : null;
```
swissiety: .

swissiety (approved):

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/DownloadJarAnalysisInputLocation.java
```diff
@@ -0,0 +1,73 @@
...
+      String downloadURL, List<BodyInterceptor> bodyInterceptors, Collection<Path> ignoredPaths) {
+    super(
+        downloadAndConstructPath(downloadURL), SourceType.Library, bodyInterceptors, ignoredPaths);
+  }
+
+  private static Path downloadAndConstructPath(String downloadURL) {
+    HttpURLConnection connection;
+    String tempDirPath = System.getProperty("java.io.tmpdir");
+    String filename = downloadURL.substring(downloadURL.lastIndexOf("/") + 1);
+    File file = new File(tempDirPath, filename);
```
marcus-h: Hmm shouldn't we use a real tmpfile? Writing to arbitrary files in the tmpdir
is not so nice:)

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/DownloadJarAnalysisInputLocation.java
```diff
@@ -0,0 +1,73 @@
...
+      try (InputStream inputStream = new BufferedInputStream(connection.getInputStream());
+          OutputStream outputStream = Files.newOutputStream(Paths.get(file.getAbsolutePath()))) {
+        byte[] buffer = new byte[BUFFER_SIZE];
+        int bytesRead;
+        while ((bytesRead = inputStream.read(buffer, 0, BUFFER_SIZE)) != -1) {
+          outputStream.write(buffer, 0, bytesRead);
+        }
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
marcus-h: Hmm maybe an UncheckedIOException?

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/DownloadJarAnalysisInputLocation.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private static Path downloadAndConstructPath(String downloadURL) {
+    HttpURLConnection connection;
+    String tempDirPath = System.getProperty("java.io.tmpdir");
+    String filename = downloadURL.substring(downloadURL.lastIndexOf("/") + 1);
+    File file = new File(tempDirPath, filename);
+    try {
+      URL url = new URL(downloadURL);
+      connection = (HttpURLConnection) url.openConnection();
```
marcus-h, line 53: Is this a potential resource leak? Do we have to manually close the connection? (I haven't checked the API...)

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/DownloadJarAnalysisInputLocation.java
palaniappan1: This is the real temp file, tmp directory by JVM.

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/DownloadJarAnalysisInputLocation.java
palaniappan1: I don't think Unchecked is really needed here. Simple file operations, so IOException itself is enough i think.

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/DownloadJarAnalysisInputLocation.java
palaniappan1: Good question. Fixed it!

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/DownloadJarAnalysisInputLocation.java
marcus-h: A user could have intentionally created a file called /tmp/foo-1.2.3.jar,
which is then overwritten by the downloader. IMHO, we shouldn't do that.
Let's use a "real" tmpfile, which is created, for instance, via
Files.createTempFile(...).

sootup.java.bytecode/src/main/java/sootup/java/bytecode/inputlocation/DownloadJarAnalysisInputLocation.java
marcus-h: I was talking about "new RuntimeException(e)" vs. "new UncheckedIOException(e)".

PR #1049, opened 2024-08-30
Refactor SootUp module structure

stschott:
- Separate sootup.analysis into sootup.analysis.interprocedural and sootup.analysis.intraprocedural
- Separate body interceptors from sootup.java.core into new module sootup.interceptors
- Move Dex interceptors into sootup.apk.parser
- Move ReachingDefs into sootup.analysis.intraprocedural
- Add new modules to report and DependencyManagement section in root pom.xml 
- add new modules to gh actions workflow ;)


sootup.analysis.intraprocedural/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.soot-oss</groupId>
+            <artifactId>sootup.java.sourcecode</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
```
JonasKlauke, line 32: here it is missing the test scope

sootup.analysis.intraprocedural/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+        </dependency>
+    </dependencies>
+    <repositories>
+        <repository>
+            <id>sonatype-snapshots</id>
```
JonasKlauke: why do we need the snapshot repo?

sootup.apk.parser/src/main/java/interceptors/AbstractNullTransformer.java
```diff
@@ -1,4 +1,4 @@
-package sootup.java.core.interceptors.Dex;
+package interceptors;
```
JonasKlauke: the package should be sootup.apk.parser.interceptors.

JonasKlauke (changes requested):

JonasKlauke:
I think we should also rename sootup.apk.parser and sootup.jimple.parser and drop the parser part

swissiety:
> I think we should also rename sootup.apk.parser and sootup.jimple.parser and drop the parser part

Then we should split the concrete Jimple Implementations from core and only keep interfaces there.

JonasKlauke (approved):

PR #1055, opened 2024-09-08
precompute toString for JavaClassType and SootClassMemberSignature

Liyw979:
like https://github.com/soot-oss/SootUp/pull/881 :)

### DIFF
--- a/sootup.core/src/main/java/sootup/core/signatures/SootClassMemberSignature.java
+++ b/sootup.core/src/main/java/sootup/core/signatures/SootClassMemberSignature.java
@@ -44,10 +44,13 @@ public abstract class SootClassMemberSignature<V extends SootClassMemberSubSigna
 
   private final int hashCode;
 
+  private final String toString;
+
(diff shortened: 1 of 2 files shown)


stschott:
Can you please describe your potential use case for this change? At the moment, in contrast to the hashcode change, we are not sure whether the trade-off in additional memory used is worth the time save.

Liyw979:
I'm doing somthing like symbolic execution, and When I put `method.getSignature()` in my code, it gets executed 1000000 times in my simulation. 😅

swissiety:
Do you actually need `toString()` there? I mean it depends on your use case but I stumbled over usage examples where `==` or `equals()` did the (better) job.

Liyw979:
Maybe we can do something like https://github.com/soot-oss/SootUp/blob/bf95782b1ec8f268d64a19748af5412031342273/sootup.core/src/main/java/sootup/core/model/SootMethod.java#L172-L178

Liyw979:
> Do you actually need `toString()` there? I mean it depends on your use case but I stumbled over usage examples where `==` or `equals()` did the (better) job.

I will try this.

Issue #1059, opened 2024-09-12
[Feature]: SootClass Builder


stschott:
### Missing Feature

Add a builder for `SootClass`, similar to `SootMethod` and `SootField`.

wkktoria:
Hi, can I work on this?

stschott:
Sure, that would be great.

PR #1069, opened 2024-09-13
Create SootClass builder

wkktoria:
As described in #1059, I've created a builder for `SootClass` which is based on `SootMethod` and `SootField`'s builders.


swissiety (approved):
thanks for your contribution! If you like to, we would really appreciate if you could do the same for the OverridingClassSource / OverridingMethodSource so SootUp can have a testcase demonstrating the usage.

wkktoria:
Sure, I'll work on making changes to these as well.

swissiety (changes requested):
Thank you! As we want to create a complete OverridingClassSource making use of multiple methods at once please return the different interfaces (as an accumulated interface/step) from the with* methods (e.g. see https://www.svlada.com/step-builder-pattern/).
(applies to both of the added Builders)

We even could think about extending the API of withMethods(Set<SootMethods> )/ withFields(Set<SootFields>) methods to singular withField(SootField) and collect the parameters on our own.

wkktoria:
I've changed the builders to use different interfaces as suggested, but I am not sure how to replace these methods with singular one.

T0niKroOs:
Thank you for the clarification! I'm looking forward to the release of the new features.

PR #1071, opened 2024-09-13
Simplified the creation of Subclasses of GraphBasedCallGraph

JonasKlauke:
Potential Subclasses gained access to edges in the call graph.
Added printCaller and printCallee to simplify the adpation of the toString method
Added toDotEdge to simplify the export of a call in the dot file generation
Introduced optionals in the graph data handling.

### DIFF
--- a/docs/analysisinput.md
+++ b/docs/analysisinput.md
@@ -115,7 +115,40 @@ Path path = Paths.get("Banana.apk");
 AnalysisInputLocation inputLocation = new ApkAnalysisInputLocation(path, "", DexBodyInterceptors.Default.bodyInterceptors());
 JavaView view = new JavaView(inputLocation);
 ```
+### Android Bytecode with Dex2Jar
+File-Extensions: `.apk`
 
+If you prefer to use dex2jar as a base to transform android apps to jimple, you can add the code below to create your own analysis input location.
+We used the dependency de.femtopedia.dex2jar:dex2jar:2.4.22 in the given example.
+We recommend to use ApkAnalysisInputLocation
+
+```java
+Path path = Paths.get("Banana.apk");
+AnalysisInputLocation inputLocation = new Dex2JarAnalysisInputLocation(path);
+JavaView view = new JavaView(inputLocation);
+
+```
+
+```java
+public class Dex2JarAnalysisInputLocation extends ArchiveBasedAnalysisInputLocation {
+
+    public Dex2JarAnalysisInputLocation(@Nonnull Path path, @Nullable SourceType srcType) {
+        super(path, srcType);
+        String jarPath = dex2jar(path);
+        this.path = Paths.get(jarPath);
+    }
+
+    private String dex2jar(Path path) {
+        String apkPath = path.toAbsolutePath().toString();
+        String outDir = "./tmp/";
+        int start = apkPath.lastIndexOf(File.separator);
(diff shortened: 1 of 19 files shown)


swissiety (commented):
seems convenient - but does this possibly enable a developer to feel confident in a way that the analysis will not fail in cases it should? e.g. thinking that it worked because there is no error and just an empty collection is returned therefore no fact exists.

JonasKlauke:
Are you saying i should print a warning?
I wasnt sure about it, it could happen often

swissiety:
I'm not happy about the warnings either :D Because their are a hint but are often times ignored anyway if its not clear why they are shown or as long as there is no supervisor who complains about them.

We have a similar situation in the `ViewTypeHierarchy` - there we have to check if the TH contains the vertex/`ClassType` - and sometimes we already know they do exist because we got the vertex from the `TypeHierarchy` and in those cases we don't need the additional check. And if its not there we throw an exception so the caller can differentiate between: vertex is unknown and vertex has none of the desired properties the method should calculate (i.e. Collections.empty() would be returned in both cases otherwise). Please check if the analogy is applicable/reasonable in this use case.

swissiety:
TLDR: is it avoidable to run into theses situations (where a log would (then) be printed)?

docs/analysisinput.md
```diff
@@ -117,42 +117,6 @@ JavaView view = new JavaView(inputLocation);
...
-        String outputFile = outDir + apkPath.substring(start + 1, end) + ".jar";
-        Dex2jarCmd.main("-f", apkPath, "-o", outputFile);
-        return outputFile;
-    }
-}
-```
-
-!!! info "A SootUp solution to directly generate Jimple is WIP!"
-
-
```
swissiety: - [ ] lets keep that code
- [ ] update the text below 
- [ ] add the needed dependency (https://github.com/soot-oss/SootUp/pull/1072)
```maven
		</dependency>
		<dependency>
			<groupId>de.femtopedia.dex2jar</groupId>
			<artifactId>dex2jar</artifactId>
			<version>2.4.22</version>
		</dependency>
```

swissiety (approved):
nice that this convenient and problem avoiding way was possible! 
please justreintroduce the doc+update ;)

swissiety (changes requested):
nice that this convenient and problem avoiding way was possible! 
please justreintroduce the doc+update ;)

docs/analysisinput.md
```diff
@@ -115,12 +115,12 @@ Path path = Paths.get("Banana.apk");
...
 ```
-
-
 ### Android Bytecode with Dex2Jar
 File-Extensions: `.apk`
 
-The `ApkAnalysisInputLocation` currently uses dex2jar internally 
+If you prefer to use dex2jar as a base to transform android apps to jimple, you can add the code below to create your own analysis input location.
+We used the dependency de.femtopedia.dex2jar:dex2jar:2.4.22 in the given example.
+We recommend to use ApkAnalysisInputLocation
```
swissiety, line 123:
```suggestion
We recommend to use `ApkAnalysisInputLocation`

```
	</dependency>
		<dependency>
			<groupId>de.femtopedia.dex2jar</groupId>
			<artifactId>dex2jar</artifactId>
			<version>2.4.22</version>
		</dependency>
```

```

swissiety (approved):

PR #1072, opened 2024-09-15
Update dex2jar

ThexXTURBOXx:
Updates dex2jar and ASM to their respective latest versions. I have squashed quite a few bugs within dex2jar since the last update :)
~~CI might be failing right now since the artifact might not have finished syncing to Central. Should work in around 15-30 minutes at most, though!~~

Edit: CI seems to have worked already, nice!

### DIFF
--- a/sootup.java.bytecode.frontend/pom.xml
+++ b/sootup.java.bytecode.frontend/pom.xml
@@ -28,17 +28,17 @@
 		<dependency>
 			<groupId>org.ow2.asm</groupId>
 			<artifactId>asm-util</artifactId>
-			<version>9.5</version>
+			<version>9.7</version>
 		</dependency>
 		<dependency>
 			<groupId>org.ow2.asm</groupId>
 			<artifactId>asm-commons</artifactId>
-			<version>9.5</version>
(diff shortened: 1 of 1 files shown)


swissiety (commented):
Thx for the update!
In the meantime we have incorporated a direct .dex to Jimple conversion so we are thinking about moving the current dependency+`AnalysisInputLocation` code to the documentation so someone can opt in into using dex2jar + additional dependency if one has the need or want to.

JonasKlauke:
Hi,
We have moved all dependency versions to the parent pom. Many of the changes were already done in the PR #1071 which happended in parallel to this PR
The dex2jar dependency was still included in the pom, but since we dont use it anymore, we removed it. 
But in the documentation we still show how to use dex2jar to create an android frontend.
But thanks for pointing out that asm.util has released version 9.7, my maven was only showing 9.6

ThexXTURBOXx:
Perfect, thanks for the news! :)
In that case, I will close this PR as it was superseded. Thank you for the heads-up!
And no problem for pointing out the ASM update ;)

Edit: And of course, let me know if there is something I can help you with!

PR #1074, opened 2024-09-15
user can pass cg, or default as CHA

### LINKED ISSUE #1064: [Feature]: Make JimpleBasedInterproceduralCFG configurable
### Missing Feature

Currently, JimpleBasedInterproceduralCFG bydefault creates a Call Graph using CHA and mainMethod signature. 
Make it configurable so that specified call graph is passed.

sahilagichani14:


swissiety (changes requested):
I'm not sure about the default:CHA..But it's convenient to explore - If we want to keep it lets inline initializeCallGraph into the ctor as it would be the only use case - and if subclassing would be needed its still achievable via the general ctor.

sootup.analysis.interprocedural/src/main/java/sootup/analysis/interprocedural/icfg/JimpleBasedInterproceduralCFG.java
```diff
@@ -109,60 +108,74 @@ public Collection<Stmt> load(SootMethod method) {
...
+      View view,
+      List<MethodSignature> entryPoints,
       boolean enableExceptions,
       boolean includeReflectiveCalls) {
     super(enableExceptions);
     this.includeReflectiveCalls = includeReflectiveCalls;
     this.view = view;
-    this.mainMethodSignature = mainMethodSignature;
-    cg = initCallGraph();
+    this.entryPoints = entryPoints;
```
sahilagichani14: @kadirayk @JonasKlauke, should we allow users to pass entry points here to create jimple cfg? Earlier I was passing entry points (which I used to create cg) instead of actual cg.getEntrypoints() (which add clinit calls also). 
This created a lot of confusion for me and was giving me NullPointerExceptions, so should we just allow cg & use entry points from there?

kadirayk: We don't need the entry points anymore if we get the CG from outside. But we should investigate the exceptions further.

sootup.analysis.interprocedural/src/main/java/sootup/analysis/interprocedural/icfg/JimpleBasedInterproceduralCFG.java
```diff
@@ -56,7 +55,7 @@ public class JimpleBasedInterproceduralCFG extends AbstractJimpleBasedICFG {
 
   protected static final Logger logger =
       LoggerFactory.getLogger(JimpleBasedInterproceduralCFG.class);
-  private final MethodSignature mainMethodSignature;
+  private List<MethodSignature> entryPoints;
```
swissiety: can't we remove that field and delegate the usages of entrypoints to cg.getEntryPoints() ?

sahilagichani14: Yes, we can, and add a getter for the call graph so one can get all info from that :)  
@swissiety assuming no one wants a incomplete jimple cfg with no clinits

swissiety (approved):

PR #1083, opened 2024-09-24
getType of JArrayRef

### LINKED ISSUE #1076: [Bug]: getType of JArrayRef is not element's type
### What happened?

test code
```
class Foo {
    public static void main(String[] args) {
        System.out.println("1" + args[0]);
    }
}
```
debugger view:
image
image

relevant code:
https://github.com/soot-oss/SootUp/blob/d6d411f6cea8b7047dc40bc86f81c518a1845f34/sootup.core/src/main/java/sootup/core/jimple/common/ref/JArrayRef.java#L84-L88

I am expecting the return type of `arrayref` to be `String` instead of `String[]`, am i right?

### Version

Latest develop branch

### Relevant log output

_No response_

Liyw979:
fix https://github.com/soot-oss/SootUp/issues/1076

### DIFF
--- a/sootup.core/src/main/java/sootup/core/jimple/common/ref/JArrayRef.java
+++ b/sootup.core/src/main/java/sootup/core/jimple/common/ref/JArrayRef.java
@@ -88,10 +88,10 @@ public final class JArrayRef implements ConcreteRef, LValue {
   @Nonnull
   public Type getType() {
     Type baseType = base.getType();
-    if (baseType.equals(NullType.getInstance())) {
-      return NullType.getInstance();
-    } else if (baseType instanceof ArrayType) {
+    if (baseType instanceof ArrayType) {
       return ((ArrayType) baseType).getElementType();
+    } else if (baseType.equals(NullType.getInstance())) {
+      return NullType.getInstance();
     } else {
       return UnknownType.getInstance();
     }
--- a/sootup.core/src/main/java/sootup/core/jimple/common/ref/JArrayRef.java
+++ b/sootup.core/src/main/java/sootup/core/jimple/common/ref/JArrayRef.java
@@ -26,7 +26,10 @@ import java.util.stream.Stream;
 import javax.annotation.Nonnull;
 import sootup.core.jimple.basic.*;
 import sootup.core.jimple.visitor.RefVisitor;
+import sootup.core.types.ArrayType;
+import sootup.core.types.NullType;
 import sootup.core.types.Type;
+import sootup.core.types.UnknownType;
 import sootup.core.util.printer.StmtPrinter;
 
 public final class JArrayRef implements ConcreteRef, LValue {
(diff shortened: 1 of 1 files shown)


sootup.core/src/main/java/sootup/core/jimple/common/ref/JArrayRef.java
```diff
@@ -84,7 +87,14 @@ public Stream<Value> getUses() {
   @Override
   @Nonnull
   public Type getType() {
-    return base.getType();
+    Type baseType = base.getType();
+    if (baseType.equals(NullType.getInstance())) {
+      return NullType.getInstance();
+    } else if (baseType instanceof ArrayType) {
+      return ((ArrayType) baseType).getElementType();
```
swissiety, line 92: I guess this if we have a multidimensional array e.g. int[ ][ ]  this would return int instead of int[ ] ? we need to reduce the amount of dimensions / unwrap as it was done in Soot.
Why they check for NullType and return UnknownType instead of the actual baseType is not clear to me (maybe sanitization but in SootUp we rather would trow an error than silently trying to fix problems introduced/not handled earlier)

swissiety (changes requested):
:)

sootup.core/src/main/java/sootup/core/jimple/common/ref/JArrayRef.java
Liyw979: `((ArrayType) baseType).getElementType();` handles the `int[ ][ ]` case I guess
https://github.com/soot-oss/SootUp/blob/621cabd9bd6dbcc898189247e5884bd8952c2003/sootup.core/src/main/java/sootup/core/types/ArrayType.java#L90-L96

Liyw979: `base.getType()` is  `UnknownType` when `TypeAssigner` is not enabled.
I feel it hard to use if we throw errors inside getType. We have to do try-catch every where.

swissiety: But baseType would already be unknown, so we can return it directly / same for NullType - Im wondering if handling just ArrayType/return baseType otherwise would be enough?

JonasKlauke: The current implementation is correct you want to have the type of a entry in an array. if the array is multidimensional you should get an array as return value.

Liyw979: I get the point. 
`else return base.getType()` is equivalent from the value perspective.
But this code would imply that the return type of jarrayref is just the array itself, this could be confusing.

JonasKlauke (approved):
I see no issue in the implementation.
Thanks 👍

sootup.core/src/main/java/sootup/core/jimple/common/ref/JArrayRef.java
JonasKlauke: The current implementation is correct you want to have the type of a entry in an array. if the array is multidimensional you should get an array as return value.

swissiety: it makes sense.. as these two would be the only valid other base types as the ArrayRef should not point anywhere else than to an Array which is of ArrayType.

swissiety (changes requested):
lets reorder the ifelse so that the ArrayType is the first check as usually someone would use the TypeAssigner

swissiety (approved):
thanks for your contribution!

PR #1089, opened 2024-09-26
move lowest common ancestor of two ClassTypes into TypeHierarchy

wangzun66:


swissiety:
@wangzun66 unfortunately something in the RuntimejarConversionTest is failing.
you can debug that faster down by adding a new testcase in that file that just executes the conversion of a single method - there should be another testcase as an example or at least in the git history of that file.

wangzun66:
> @wangzun66 unfortunately something in the RuntimejarConversionTest is failing. you can debug that faster down by adding a new testcase in that file that just executes the conversion of a single method - there should be another testcase as an example or at least in the git history of that file.

OK 👌

swissiety (approved):

PR #1090, opened 2024-09-27
make use of IdentityHashMap in JimpleICFG

swissiety:


sootup.analysis.interprocedural/src/main/java/sootup/analysis/interprocedural/icfg/AbstractJimpleBasedICFG.java
```diff
@@ -107,8 +112,8 @@ public List<Stmt> getSuccsOf(Stmt stmt) {
     if (body == null) {
       return Collections.emptyList();
     }
-    StmtGraph<?> unitGraph = getOrCreateStmtGraph(body);
-    return unitGraph.successors(stmt);
+    StmtGraph<?> stmtGraph = getOrCreateStmtGraph(body);
```
stschott, line 115: Can we be consistent and call it either `stmtGraph` or `cfg` across the whole class?

stschott (approved):

PR #1092, opened 2024-09-30
Adding support for backwards interprocedural CFG

shazib78:
Adding Backwards interprocedural CFG to allow backward analysis.
-The "BackwardsInterproceduralCFG" class has been taken from Soot, and made compatible with SootUP by changing the references of Unit to Stmt and DirectedGraph to StmtGraph.


swissiety (approved):
thx for your contribution!

PR #1107, opened 2024-10-13
add HasAnnotation for JavaSootClass, JavaSootField and JavaSootMethod

Liyw979:
like `sootup.core.model.HasPosition`

In the follwing case, I want to get urls from class(/fruit) and method (/lychee) and get the full url.

```
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RequestMapping;
import org.springframework.web.bind.annotation.RestController;

@RestController
@RequestMapping("/fruits")
public class LycheeController {

    @GetMapping("/lychee")
    public String getLychee() {
        return "Here is a fresh lychee!";
    }
}
```
by adding the HasPosition interface, users could write functions like 
```
String getUrl(has: HasAnnotation) { // "has" could be JavaSootClass or JavaSootMethod
    has.getAnnotations .....
}
```


swissiety (approved):
thx!

PR #1121, opened 2024-10-25
Fix/default runtime analysis input location

stschott:
Fixed a bug where the wrong RtInputLocation is selected for Java versions 10-19.


swissiety (approved):
good **point** !

PR #1125, opened 2024-11-01
add failing test case

### LINKED ISSUE #1119: [Refactoring]: Move the StmtGraph.buildTraps() API into the JimplePrinter
### Missing Feature

The StmtGraph.buildTraps() API should only be used for the serialization for Jimple and should therefore be moved to the JimplePrinter. This also requires the adjustment of some code depending on this API (e.g. Body.getTraps() and uses in test cases).

Additionally, to avoid the "invalid StmtGraph because a trap's end stmt was not visited" exception when printing traps, add a nop statement in cases where a trap's end stmt is missing.

sahilagichani14:


sootup.core/src/main/java/sootup/core/util/printer/BlockGraphIteratorAndTrapAggregator.java
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Iterates over the Blocks and collects/aggregates Trap information It is used to collect and
+ * aggregate traps for serializing Jimple in the JimplePrinter
+ */
+class BlockGraphIteratorAndTrapAggregator extends BlockGraphIterator {
+
+  @Nonnull private final List<Trap> collectedTraps = new ArrayList<>();
+
+  Map<ClassType, Stmt> activeTraps = new HashMap<>();
+  BasicBlock<?> lastIteratedBlock; // dummy value to remove n-1 unnecessary null-checks
```
swissiety: lets remove that comment

sootup.core/src/main/java/sootup/core/util/printer/LabeledStmtPrinter.java
```diff
@@ -105,16 +107,15 @@ public void stmtRef(Stmt stmt, boolean branchTarget) {
...
+    final List<Stmt> linearizedStmtGraph = getStmts(stmtGraph, traps);
     return linearizedStmtGraph;
   }
 
   @Nonnull
-  public List<Stmt> getStmts(@Nonnull StmtGraph<?> stmtGraph) {
-    final Collection<Stmt> targetStmtsOfBranches = stmtGraph.getLabeledStmts();
-    final List<Trap> traps = stmtGraph.buildTraps();
+  public List<Stmt> getStmts(@Nonnull StmtGraph<?> stmtGraph, List<Trap> traps) {
+    final Collection<Stmt> targetStmtsOfBranches = getLabeledStmts(stmtGraph, traps);
```
swissiety: lets rename targetStmtsOfBranches to labeledStmts

sootup.core/src/main/java/sootup/core/util/printer/LabeledStmtPrinter.java
```diff
@@ -163,6 +164,15 @@ public List<Stmt> getStmts(@Nonnull StmtGraph<?> stmtGraph) {
...
     }
+
+    // add Nop Stmt to Jimple just for serialization
+    for (Stmt s : targetStmtsOfBranches) {
+      if (trapStmts.contains(s) && s instanceof JNopStmt) {
+        linearizedStmtGraph.add(s);
+        labels.put(s, String.format(formatString, ++labelCount));
+      }
+    }
+
```
swissiety: seems complicated - Isn't it only necessary to add a `JNopStmt` as the last stmt (i.e. once)?
and it seems this JNopStmt would not be added to labelStmts in line 141

swissiety (changes requested):

sootup.core/src/main/java/sootup/core/util/printer/LabeledStmtPrinter.java
sahilagichani14: Since we have separate trapStmts in which we already added nop stmt (when we create a nop stmt). To add that same nop stmt as a separate labeledStmt, I'm trying to find that specific nop stmt in targetStmtsOfBranches. Because in one of the test case there were couple of Nop stmts which we can't distinguish so that's the reason I added a for loop and see nop stmt which is in trapStmts.

swissiety: at least from my understanding we only need to insert a JNopStmt once - even when multiple `Trap` objects would reference it as this additional Nop is only needed as a final Stmt - in the other scenarios there is a following Block that can be serialized after the current Block and therefore we can reference that following blocks headStmt.

sahilagichani14: Yes the Nop stmt added at last is the only inserted JNopStmt once - even when multiple Trap objects would reference it as this additional Nop is only needed as a final Stmt. But if StmtGraph already contains nop stmts then to distinguish  those with this trap nop stmt I'm checking only in trapStmts

sootup.core/src/main/java/sootup/core/graph/StmtGraph.java
```diff
@@ -65,6 +66,9 @@
  */
 public abstract class StmtGraph<V extends BasicBlock<V>> implements Iterable<Stmt> {
 
+  public final JimplePrinter jimplePrinter = new JimplePrinter();
+  public final BriefStmtPrinter briefStmtPrinter = new BriefStmtPrinter();
```
swissiety: please dont

swissiety (approved):

sootup.core/src/main/java/sootup/core/graph/StmtGraph.java
sahilagichani14: Idk why it's still there, There's just one in toString new JimplePrinter().printTo(this, writerOut);

sahilagichani14: removed fields and used them directly

PR #1126, opened 2024-11-05
Fix switch comparison

stschott:


swissiety (approved):
:+1:

swissiety:
hmm... `MutableBlockStmtGraphTest.testRemoveMultipleTrapsWithDifferentExceptionTypes:1098 » IllegalArgument Invalid StmtGraph. A Trap is not created as a traps endStmt was not visited during the iteration of all Stmts.`

stschott:
Flaky test? Locally the test case passes.

stschott:
The only method I changed is not even called when running the `MutableBlockStmtGraphTest.testRemoveMultipleTrapsWithDifferentExceptionTypes` test.

PR #1128, opened 2024-11-06
coupled two method streams into single stream in APK frontend

palaniappan1:


swissiety (approved):
I get the theme - currently its all about tying entities together :ring:

PR #1141, opened 2024-11-28
fix insertBefore / implement insertAfter

### LINKED ISSUE #751: implement MutableStmtGraph.insertAfter(..)
`MutableStmtGraph.insertAfter(FallsThroughStmt|BranchingStmt afterStmt, FallsThroughStmt newStmt)`

### LINKED ISSUE #873: fix (semantics) of MutableStmtGraph.insertBefore(Stmt existingStmt, FallsThroughStmt newStmt)
https://github.com/soot-oss/SootUp/pull/869/files#diff-c5721a80c02fabaa342480b1e4d00ff3269aa3bc5ca20c721c4d30c0ad3fac75R87
suggests that something is wrong with insertBefore(...).

semantic options:
a) All (Branching-) predecessors of existingStmt would flow to the newStmt and then it goes to the existingStmt. 
b) inserts a newStmt in the FallsThrough flow but not in the predecessor flows branching to the oldStmt.

- [ ] it seems a) is more intuitive for insertBefore(...)
- [ ] and an easier option for b) than: removeEdge(...), putEdge(...) etc. should exist as well.

wangzun66:
* insertBefore
- [x] insert a stmt before a block head stmt with same trap-map
- [x] insert a stmt before a block head stmt with different trap-map
- [x] insert a stmt before a block middle/tail stmt with same trap-map
- [x] insert a stmt before a block middle/tail stmt with different trap-map
* insertAfter
- [x] insert a stmt after a block middle/head stmt with same trap-map
- [x] insert a stmt after a block middle/head stmt with different trap-map
- [x] insert a stmt after a block tail stmt with same trap-map
- [x] insert a stmt after a block tail stmt with different trap-map
* other problems
- [x] P1: After insertion, Blocks are not merged.
- [x] P2: After insertion,  Information of Traps may be false.
- [x] P3: Sometimes(maybe with traps), a block could have two same predecessors.


sootup.core/src/main/java/sootup/core/graph/MutableStmtGraph.java
```diff
@@ -67,17 +67,27 @@ public void addBlock(@Nonnull List<Stmt> stmts) {
...
    */
-  public void insertBefore(@Nonnull Stmt beforeStmt, @Nonnull FallsThroughStmt newStmt) {
-    insertBefore(beforeStmt, Collections.singletonList(newStmt), Collections.emptyMap());
+  public BasicBlock<?> insertBefore(@Nonnull Stmt beforeStmt, @Nonnull FallsThroughStmt newStmt) {
+    return insertBefore(beforeStmt, Collections.singletonList(newStmt), Collections.emptyMap());
+  }
+
+  /** inserts the "newStmt" after the position of "afterStmt" */
+  public BasicBlock<?> insertAfter(@Nonnull Stmt afterStmt, @Nonnull FallsThroughStmt newStmt) {
+    return insertAfter(afterStmt, Collections.singletonList(newStmt), Collections.emptyMap());
```
swissiety, line 90: which behaviour do we want:
a) no exception for the new stmt (current)
b) fit in seamlessly  into existing block/surrounding structure (of the anchoring Stmt i.e. afterStmt)

wangzun66: Behavior a. That means the new stmt is not in a trap.  If the anchoring stmt is also not in a  trap, then the behavior is just like b. If the anchoring stmt is in a trap, the new stmt and anchoring stmt will be in two different blocks.

wangzun66: PS: The same thing happens with insertBefore.

swissiety (approved):

PR #1145, opened 2024-12-07
remove Optional since buildClassFrom always finds a class

Liyw979:


swissiety (approved):
great job! can you check if the other Views (JavaView,JimpleView) have the same logic?

PR #1158, opened 2025-01-12
[code improve] No more views passing around

Liyw979:
since We have access to the current view object  in 
https://github.com/soot-oss/SootUp/blob/76f7e78f2f1a0103b4ac3be448c553fb6184bec5/sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java#L62-L64


swissiety (commented):
makes sense

PR #1162, opened 2025-01-20
Jimple to JavaObjects to change minimaltestsuite

### LINKED ISSUE #1056: [Feature]: JimpleJavaObjectPrinter
### Missing Feature

usecase: simplifying testcases

- [ ] replace minimaltestsuite tests by its equivalent output of the JimpleJavaObjectPrinter.
- [ ] move testcases into core module / remove then unnecessary dependency to bytecodefronted
- [ ] keep some testcases - especially the ones with try-resources, if-else, try-catch, synchronized, ... as JimplePrinterTests

sahilagichani14:


swissiety (changes requested):
Just to be clear we are on the same page: the goal is to transform a StmtGraph to a String containing Java Sourcecode. And if we run that Javacode we should get the equivalent StmtGraph (again)

sahilagichani14:
> Just to be clear we are on the same page: the goal is to transform a StmtGraph to a String containing Java Sourcecode. And if we run that Javacode we should get the equivalent StmtGraph (again)

So the starting point is StmtGraph and not Jimple String?

swissiety:
yes and please make use/implement a `StmtVisitor` to print the details of the respective Stmts. And in those methods of the `StmtVisitor` you could make use of that JavaPoet library if it helps you to keep track of the different Java variables of the generated source code. Maybe a `Map<Value, JavaVariableNameString>` of already generated Jimple objects would suffice as well, so you could generate each Stmt and initialize its Operands or reuse the already created ones.

sootup.jimple.frontend/src/main/java/sootup/jimple/frontend/JavaCodeBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+              ((Local) stmt.getRightOp()).getName(), stmt.getRightOp().getType());
+    }
+    javaCodeObjects.add(
+        String.format(
+            "Stmt %s = JavaJimple.newAssignment(JavaJimple.newLocal(\"%s\", factory.classType(\"%s\")), %s , noStmtPositionInfo)",
+            stmt.getLeftOp().toString(),
+            stmt.getLeftOp().toString(),
+            stmt.getType().toString(),
+            typeString));
+  }
```
swissiety: don't forget we have more of those visitors like the ValueVisitor e.g. for Operands

sootup.java.core/src/main/java/sootup/java/core/buildsrccode/JavaSrcCodeBuilder.java
```diff
@@ -0,0 +1,234 @@
...
+
+  public void createStmtGraph(Body body) {
+    javaCodeObjects.add("MutableStmtGraph stmtGraph = bodyBuilder.getStmtGraph();");
+    javaCodeObjects.add(
+        String.format("stmtGraph.setStartingStmt(%s);", stmtVarName.get(body.getThisStmt())));
+    // loop over statements and put edges
+    List<Stmt> bodyStmts = body.getStmts();
+    if (!bodyStmts.isEmpty()) {
+      Iterator<Stmt> stmtIterator = bodyStmts.iterator();
+      // TODO: how to add edges or blocks when jumping statements?
```
sahilagichani14: @swissiety can we have a call to discuss this part?

swissiety (approved):

PR #1165, opened 2025-01-23
Remove duplicated null and body checks in  AbstractCallGraphAlgorithm

Liyw979:


Liyw979:
Look like there is something wrong with the CI ...

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
```

JonasKlauke:
I get that we can save time by doing the check before, but also these Methods have visibility protected. Shouldnt we provide robust methods as a framework?

JonasKlauke:
The null check can be saved if we annote the parameter with nonnull. I'm not sure if we should remove the body check

swissiety:
we can kill these nullchecks here as with `@Nonnull` parameters they just hide bugs originating from somewhere else. Especially when these methods are not public api we should be able to control the originating parts.
Would keep the hasBody unless we make sure even isAbstract() and isNative() will refer to a default/emptyBody.

Liyw979:
Now we kill null checks using `@Nonnull` and `ifPresent` and keep `if (!sourceMethod.hasBody()) {`

Liyw979:
Why the CI `publish-javadoc`  is failing...
```
/usr/bin/chmod -R 777 github-pages-deploy-action-temp-deployment-folder
/usr/bin/git worktree remove github-pages-deploy-action-temp-deployment-folder --force
Error: The deploy step encountered an error: The process '/usr/bin/git' failed with exit code 128 ❌
Deployment failed! ❌

Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

JonasKlauke:
this is adressed in #1398

PR #1167, opened 2025-01-23
Simplify code in RapidTypeAnalysisAlgorithm

Liyw979:
Use ArrayListMultimap and We don't need to worry about nulls


sootup.callgraph/src/main/java/sootup/callgraph/RapidTypeAnalysisAlgorithm.java
```diff
@@ -49,7 +50,7 @@
 public class RapidTypeAnalysisAlgorithm extends AbstractCallGraphAlgorithm {
 
   @Nonnull protected Set<ClassType> instantiatedClasses = Collections.emptySet();
-  @Nonnull protected Map<ClassType, List<Call>> ignoredCalls = Collections.emptyMap();
+  @Nonnull protected ArrayListMultimap<ClassType, Call> ignoredCalls = ArrayListMultimap.create();
```
swissiety, line 53: can we make them final and initialize them just in the constructor?

swissiety (approved):

sootup.callgraph/src/main/java/sootup/callgraph/RapidTypeAnalysisAlgorithm.java
Liyw979: Yes, I have the same thought.

JonasKlauke: it isnt final since it is a big data structure which is only needed in the creation. The Map is thrown away after the processing.
I guess if we want to make it final we have to clear the Map.

Liyw979: Yes, we can replace `ignoredCalls = ...` with  `ignoredCalls.clear()` and make it final.

JonasKlauke: I solved the issue of RTA only using initialized Classes and the On the fly creation by creating a ignored calls Map, that saves possible edges for a specific class, if the algorithm ever sees an initilaiziation of the class all possible edges are taken from the Map and the entry is cleared.
So we can still process every method only once and still adapt the edges to later found information.
But This approach is quite memory heavy, so I clear the Map after the algorithm is done.

swissiety: is it necessary to have that temporary map as a field or does make sense if we have some kind of (temporary) builder class that stores that information etc.? we could move CallGraph.addCall there as well

PR #1172, opened 2025-01-28
Use canonical dest dir path in ZipSlip check

eaftan:
I was attempting to build this project on my Mac, and one test -- `PathBasedAnalysisInputLocationTest#testWar` -- failed.  It seems that at least on my Mac, the destination directory is not canonical, and so the check here incorrectly fails.  The test passes once I canonicalize the destination directory.


swissiety (approved):
:+1: thanks for the contribution!

PR #1174, opened 2025-01-31
Change log level in the dominance finder due to too many logs

stschott:


swissiety (approved):

swissiety:
> [Documentation Preview](https://soot-oss.github.io/SootUp/fix_changeLogLevel_preview/).

wow. sidenote: so the path filter in the gh-workflow triggered it from the merge commit and not from changed files..

PR #1176, opened 2025-02-04
Fix Exception Flow Control - Update Trap Comparison (Issue 1175)

### LINKED ISSUE #1175: [Bug]: Incorrect Exception Control Flow Analysis for Nested Try-Catch Blocks (1.3.0)
### What happened?

When analyzing nested try-catch blocks, SootUp incorrectly generates exceptional successor blocks that violate Java's exception handling semantics. Specifically, it suggests that a single exception throw can reach an outer unreachable catch block immediately without first going through the first catch block of the same exception type.

Steps to reproduce:

Create a Java method with nested try-catch blocks containing multiple catches for the same exception type:
```java
    public int test_nested_try_catch_2(int param) {
        try {
            try {
                try {
                    if (param == 10)
                        return 0;
                    if (param == 20 || param == 25 || param == 27 || param == 28)
                        throw new IllegalArgumentException("Illegal arguments provided");
                    if (param == 30)
                        throw new IllegalStateException("Illegal state exception");
                    return 1;
                } catch (IllegalArgumentException e) {
                    if (param == 25 || param == 27 || param == 28)
                        throw new IllegalStateException("Illegal state exception 2");
                    return 2;
                } catch (IllegalStateException e) {
                    return 3;
                }
            } catch (IllegalStateException e) {
                if (param == 25)
                    throw new ArrayIndexOutOfBoundsException();
                if (param == 27)
                    throw new IllegalArgumentException("Illegal arguments provided 2");
                return 4;
            }
        } catch (ArrayIndexOutOfBoundsException e) {
            return 5;
        } catch (RuntimeException e) {
            return 6;
        }
    }
```

Analyze the method using SootUp:
Expected behavior:
When an IllegalStateException is thrown in the innermost try block, it should only be able to reach the first matching catch block (which returns 3).

Actual behavior:
Within the inner try block only the outer IllegalStateException is reachable, skipping the first IllegalStateException catch block completely.

Here is the image of the exception successor blocks:
Image

I am working on a symbolic executor and this issue is vital to me.

### Version

Latest develop branch

### Relevant log output

```shell

```

A1K28:
- [x] add test for input
Closes https://github.com/soot-oss/SootUp/issues/1175


swissiety (changes requested):
lets add a test (input is already pushed) that checks that the correct exceptional successor is linked. for simplicity sth like graph.getBlocks().stream().filter(b -> b.toString().contains("return o")).findFirst().get().getExceptionalSuccessors().values().stream().filter(ehb -> ehb.toString().contains("return 2")).findAny() == true; (beware: its possible the exceptionhandler consists of multiple blocks)

A1K28:
@swissiety I've added the test as you requested. It's not completely generic, but it should suffice. Let me know if you want me to change anything.

swissiety:
Wow thats a lot of assertions. Thank you for the testcase!

sootup.tests/src/test/java/sootup/tests/NestedTryCatchFlowTest.java
```diff
@@ -0,0 +1,193 @@
...
+import sootup.java.core.views.JavaView;
+
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class NestedTryCatchFlowTest {
```
swissiety, line 30: Please annotate it with the test category tag (see other tests with java8 something something)

swissiety:
Please execute the coveo plugin with the fmt goal to apply the code style before committing

A1K28:
i think i did. not sure if that's what you wanted

swissiety (approved):
Good work @A1K28 :+1:

PR #1181, opened 2025-02-18
UCE block approach, & 2 tests

### LINKED ISSUE #930: [Bug]: ConditionalBranchfolder removes complete tail of the CFG
### What happened?

not even a  JReturn*Stmt anymore.

see RuntimeJarConversionTests

### Version

Latest develop branch

### Relevant log output

_No response_

sahilagichani14:


sootup.interceptors/src/main/java/sootup/interceptors/UnreachableCodeEliminator.java
```diff
@@ -21,58 +21,68 @@
...
-      graph.removeNode(stmt, false);
-      builder.removeDefLocalsOf(stmt);
+    List<? extends BasicBlock<?>> unreachableBlocks =
+        allBlocks.stream()
+            .filter(basicBlock -> !reachableNodes.contains(basicBlock))
+            .collect(Collectors.toList());
+
+    // remove block from blocks
+    for (BasicBlock<?> unreachableBlock : unreachableBlocks) {
+      graph.removeBlock(unreachableBlock);
```
stschott: Can you combine these two steps into one? No need to collect to a list. Just use stream().filter(...).forEach(...);

stschott (changes requested):

sootup.interceptors/src/main/java/sootup/interceptors/UnreachableCodeEliminator.java
sahilagichani14: Actually I tried that but it gives java.util.ConcurrentModificationException for some cases where, hence I choose this and it's better for understanding as well. But I used foreach instead of loop

stschott: Then use an iterator directly: https://www.baeldung.com/java-concurrentmodificationexception
We want to avoid having to iterate the list twice.

stschott (approved):

PR #1192, opened 2025-03-06
Fixed missing resources in call graph test

JonasKlauke:


JonasKlauke:
@sahilagichani14 since all PR dont fail. could it be that this test is not tested?
Is the test needed? I found this issue by running all testcases

JonasKlauke:
this PR is contained in  #1195

PR #1196, opened 2025-03-10
add constructor check utility methods to SootClass

stschott:


sootup.core/src/main/java/sootup/core/model/SootMethod.java
```diff
@@ -220,6 +220,16 @@ public boolean isMain(@Nonnull IdentifierFactory idf) {
...
+  /** @return true if the method is a constructor */
+  public boolean isConstructor(@Nonnull IdentifierFactory idf) {
+    return idf.isConstructorSignature(getSignature());
+  }
+
+  /** @return true if the method is the default constructor */
+  public boolean isDefaultConstructor(@Nonnull IdentifierFactory idf) {
+    return isConstructor(idf) && getParameterCount() == 0;
+  }
+
```
swissiety, line 232: shall we move that into the IdentifierFactory and pass a Signature?

stschott: The IDF already contains the methods for checking singatures. I think it's more intuitive to look for these checks on the `SootMethod` itself. All the modifier checks are performed on the `SootMethod`. It seems kinda unintuitive and not obvious for unexperienced users to have to look at the IDF for this info. We might even consider to have a more concrete implementation on the `JavaSootMethod` that does not even need the IDF as parameter, but directly uses the Java IDF.

kadirayk (approved):

Issue #322, opened 2020-11-09
converting/applying old Soots Validators


swissiety:
work through folder and convert old Validators (see `de.upb.swt.soot.core/src/main/java/de/upb/swt/soot/core/validation/*`) to current:
1. use gathererd information from there to improve doc for Jimple
2. improve the speed of "getter" methods (getThisLocal, getParameterLocal etc - e.g. dont iterate&compare! access/check index position where possible)

PR #1277, opened 2025-04-18
794 check for body validations and class validations

### LINKED ISSUE #794: Check for valid modifier combinations
FieldModifier.toString(...) already assumes valid Modifier combinations (theoretically its currently possible to assign public+private+protected modifier at the same time) - maybe we should implement a wrapper of EnumSet<...Modifier> to handle only valid combinations and implement toString there?

_Originally posted by @swissiety in https://github.com/soot-oss/SootUp/pull/793#discussion_r1440506355_

sahilagichani14:


sootup.core/src/main/java/sootup/core/model/FieldModifier.java
```diff
@@ -133,12 +131,6 @@ public static String toString(@NonNull Set<FieldModifier> m) {
     return builder.toString();
   }
 
-  @NonNull
-  // depends on the natural order of the Enums!
```
sahilagichani14, line 137: removed as no usage, and in which case we don't need to print in order?

sootup.core/src/main/java/sootup/core/validation/FieldModifiersValidator.java
```diff
@@ -22,13 +22,81 @@
...
+import java.util.Arrays;
+import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
+import sootup.core.model.FieldModifier;
 import sootup.core.model.SootClass;
 import sootup.core.model.SootField;
 
 /** Validator that checks for impossible combinations of field modifiers */
 public class FieldModifiersValidator implements ClassValidator {
```
sahilagichani14, line 35: I reused this class, which was already there, or should I create another separate class for this wrapper?

stschott: This is linked to #322 
You can directly address the other validators as well.

swissiety (approved):

PR #1278, opened 2025-04-20
make all methods Jimple.newXYZ() static

### LINKED ISSUE #196: Fix inconsistency between static and non-static Jimple factory methods
The classes `Jimple` and `JavaJimple` serve as factories for Jimple entities for easy discoverability and minor simplification. As the name implies, `JavaJimple` contains factory methods for Java-specific behavior.

Some of these methods are static and some are non-static, which is inconsistent. One reason is that `Jimple` contains some abstract methods and `JavaJimple` is in the Java Core module, making it impossible to create an instance from within the Soot Core module.

One way to solve this problem would be to

1. Eliminate calls to `Jimple.newXYZ(...)` within the Soot Core module by inlining them.
2. Move all logic from `Jimple` to `JavaJimple` as static methods. Users then don't have to think about whether they should look for a method in `Jimple` or `JavaJimple`, as there will only be one.
3. When adding more languages in the future, create new factory classes like `PythonJimple`

Alternative approaches are also welcome.

sahilagichani14:


stschott:
I think it is better if we make all methods in `Jimple` and `JavaJimple` static and remove the whole `getInstance()` part. Instead we should just call these methods statically. Also it makes sense to move the `Jimple.escape()` and `Jimple.unescape()` methods into a separate utility.

sootup.java.core/src/main/java/sootup/java/core/language/JavaJimple.java
```diff
@@ -70,35 +63,37 @@ public static JavaLocal newLocal(String name, Type t, Iterable<AnnotationUsage>
     return new JavaLocal(name, t, annotations);
   }
 
-  public JCaughtExceptionRef newCaughtExceptionRef() {
+  /** Constructs a CaughtExceptionRef() grammar chunk. */
+  public static JCaughtExceptionRef newCaughtExceptionRef() {
     return new JCaughtExceptionRef(getIdentifierFactory().getType("java.lang.Throwable"));
```
stschott: Please create the `JavaIdentifierFactory` as a static final field and unify its usage across the whole `JavaJimple` class.

stschott (changes requested):

stschott (approved):

PR #1297, opened 2025-05-19
Sorting becomes optional in the callgraph dot exporter or can be configured

JonasKlauke:
Introduces unsorted dot export
Introduces sorted dot export by a given Comparator
Introduces 2 preset sortings

### DIFF
--- a/sootup.callgraph/src/main/java/sootup/callgraph/CallGraph.java
+++ b/sootup.callgraph/src/main/java/sootup/callgraph/CallGraph.java
@@ -22,11 +22,10 @@ package sootup.callgraph;
  * #L%
  */
 
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Stream;
 import org.jspecify.annotations.NonNull;
 import sootup.core.jimple.common.stmt.InvokableStmt;
 import sootup.core.signatures.MethodSignature;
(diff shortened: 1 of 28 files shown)


sootup.callgraph/src/main/java/sootup/callgraph/CallGraph.java
```diff
@@ -180,8 +191,101 @@ boolean containsCall(
...
+            // src parameter list
+            .thenComparing(call -> call.getSourceMethodSignature().getParameterTypes().toString())
+            // target class name
+            .thenComparing(
+                call -> call.getTargetMethodSignature().getDeclClassType().getClassName())
+            // target method name
+            .thenComparing(call -> call.getTargetMethodSignature().getName())
+            // target parameter list
+            .thenComparing(call -> call.getTargetMethodSignature().getParameterTypes().toString()));
+  }
```
swissiety: shall we provide just these Comparators + metntion them in the Javadoc to not clutter the interface?

swissiety (approved):

sootup.callgraph/src/main/java/sootup/callgraph/CallGraph.java
marcus-h: A related question: Should the export* etc. methods be part of the CallGraph interface at all? What about providing a separate DotExporter class instead?

(Unrelated side note: Hmm is building/returning a "big" in-memory String a good idea, in case of a "large" callgraph?)

JonasKlauke: I can change it to a String List, but if we dont want to handle files I dont see a memory saving other way. 
Also this method can be build by the user themselves using all interface methods, in case of a more optimized version, the user can build it themselves

sootup.qilin/src/main/java/qilin/core/builder/callgraph/OnFlyCallGraph.java
```diff
@@ -483,8 +490,8 @@ public boolean containsCall(
 
   @Override
   public boolean containsCall(@NonNull Call call) {
-    if (this.calls.containsKey(call.getSourceMethodSignature())) {
-      return this.calls.get(call.getSourceMethodSignature()).contains(call);
+    if (this.calls.containsKey(call.sourceMethodSignature())) {
+      return this.calls.get(call.sourceMethodSignature()).contains(call);
```
swissiety: lets combine this.calls.get/containskey -> != null

swissiety (approved):

PR #1304, opened 2025-05-26
Simplify the callgraph algorithm to speed up

### LINKED ISSUE #1194: [Bug]: the method called in a special invoke can be different than the signature
### What happened?

If a super operator is used the method invoke will be a special invoke, but compared to other special invokes, like private methods and constructor calls, the super operator can be used to call any method of a super class.
This target must be also resolved since the actually implementation of the method can be in a supersuperclass.

Small example: 
testcase NVC 4

```java
class Class extends Superclass {

//  @DirectCall(name = "method", line = 9, resolvedTargets = "Lnvc/Rootclass;")
  protected void method(){
    super.method();
  }

  public static void main(String[] args){
    Class cls = new Class();
    cls.method();
  }
}

class Superclass extends Rootclass {

}

class Rootclass {
  protected void method(){ /* do something relevant */ }
}
```

### Version

Latest develop branch

### Relevant log output

```shell

```

JonasKlauke:
simplified the algorithm to call less the expensive method that calculates the concrete dispatch.
Default method handling is only done when the base method is a default method.
Fixed #1194 that did not correctly resolve special invokes


sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -556,4 +543,87 @@ public static Optional<SootMethod> findConcreteMethod(
...
+        if (defaultMethod == null) {
+          defaultMethod = fittingDefaultMethod;
+          continue;
+        }
+        // is the same interface method
+        if (fittingDefaultMethod.getSignature().equals(defaultMethod.getSignature())) {
+          continue;
+        }
+        // is subtype of the current default method
+        if (view.getTypeHierarchy()
```
swissiety: lets cache that getTypehiaerarchy call

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -556,4 +543,87 @@ public static Optional<SootMethod> findConcreteMethod(
...
+                        .getMethod(
+                            view.getIdentifierFactory()
+                                .getMethodSignature(classType, defaultSignature))
+                        .stream())
+            .toList();
+
+    SootMethod defaultMethod = null;
+    // find default method
+    if (!fittingDefaultMethods.isEmpty()) {
+      for (SootMethod fittingDefaultMethod : fittingDefaultMethods) {
```
swissiety: instead of having an additional loop after collecting everything into a list: lets put that into a [reduce](https://www.baeldung.com/java-stream-reduce) call?

swissiety (approved):

PR #1325, opened 2025-06-13
fix: Make type visitor pass ArrayType on case.

user-L9b40P:
Allow for type visitors to have access to the underlying `ArrayType` instance on `caseArrayType` case.

As requested in this question [here](https://github.com/soot-oss/SootUp/discussions/1199#discussion-8064956) it would be useful for accessing type information when building a method descriptor.


swissiety (approved):
@dload0 thx for your contribution!

swissiety:
merge in another PR

PR #1328, opened 2025-06-21
replace jgrapht with graph4j in PAG and CallGraph

### LINKED ISSUE #1316: [Feature]: replace the backing graph library of TypeHierarchy and CallGraph by a more performant choice
JGraphT seeems to be quite slow - see https://arxiv.org/pdf/2308.09920

https://central.sonatype.com/artifact/org.graph4j/graph4j

sahilagichani14:


sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -75,15 +79,15 @@ protected GraphBasedCallGraph(
...
+    Vertex v = new Vertex(calledMethod, vertexIdCounter++);
     addMethod(calledMethod, v);
   }
 
   protected void addMethod(@NonNull MethodSignature calledMethod, Vertex vertex) {
     if (containsMethod(calledMethod)) {
       return;
     }
-    graph.addVertex(vertex);
+    graph.addLabeledVertex(vertex);
```
swissiety: correct me if I'm wrong - but can't we pass in the MethodSignature itself? so we don't even need the Vertex class, anymore.

swissiety (approved):

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
sahilagichani14: Hey, to add custom classes, we can either add an int (id for each vertex) and a label (Vertex class). Another option was to have 2 HashMaps to map IDs with method signatures, as I couldn't find an API to get a Mapping between vertex ID and label.

sahilagichani14: @swissiety, one question though, for the graph4J lib, it supports only int/primitive types for edges and vertices. Therefore, if I map the method signature with a unique int, it wouldn't be a good approach and might even exceed the int max value. I just want to ask before switching to a different library.

Also, the tests that fail in the pipeline are green locally, idk whats wrong in the pipeline?

swissiety:
I think would make sense to me to make use of a custom datastructure. We don't really benefit (quite the contrary: performance wise we sometimes iterate over all saved/joined neighbouring edges of a vertice-> O(neighbors) to find a single extends) from a general graph datastructure as we are basically just storing multiple trees within the same graph which has the same vertices in common.

swissiety:
@sahilagichani14 maybe it makes sense to extend existing graph4j structures to store those trees? especially making use of storing primitive integer values representing the vertices like graph4j does so that we can later on implement an efficient version of LCA (used in type inference) without the need for additional mappings etc - for lca see https://www.baeldung.com/cs/tree-lowest-common-ancestor

swissiety:
@Sahil lets merge this branch: i.e. please revert the failing changes from the typehierarchy - lets tackle that in another PR

JonasKlauke:
The graph is not optimized and implemented to support Pseudograph.
The API does not support to return multiple edges betwenn the same vertexes.
Additonally, jgrapht has a bigger community. Graph4J is run by 2 developers.
I'm not sure if we should change. 
I think the overhead that is caused to use graph4j with multiple edges is bigger than the gain.
I would say we shoudl put this PR as not planned

JonasKlauke (changes requested):
multi edges support needed in graph4j

PR #1331, opened 2025-06-24
Fix cite of the repository

JonasKlauke:


swissiety (approved):

swissiety:
it seems we need a json validation workflow on the citation file?! :laughing:

PR #1332, opened 2025-06-25
FixJars Test Fixed

### LINKED ISSUE #1321: [Feature]: please handle test random jars from differently
### Missing Feature

its quite annoying to commit etc. after testing locally when FixJars.java gets replaced/created all the time

palaniappan1:
only if there are any failed test information it will be written to a test file


palaniappan1:
#1321

sootup.java.bytecode.frontend/src/test/java/sootup/java/bytecode/frontend/inputlocation/TestWriter.java
```diff
@@ -47,7 +47,12 @@ public String getMethodString(String jarDownloadUrl, String methodSignature) {
 
   public String getTestContent() {
     StringBuilder content = new StringBuilder(templateStart);
-    for (JarFailureRecord record : getRecords()) {
+    List<JarFailureRecord> records = getRecords();
+    if (records.isEmpty()) {
+      System.out.println("No failed jars information found, so nothing to write");
```
JonasKlauke: no sysout in tests. only report errors

JonasKlauke:
This tests probably also creates a FixJar file
https://github.com/soot-oss/SootUp/blob/b6f1aac5a45be1eb0264ac400bd59fdead7809e7/sootup.java.bytecode.frontend/src/test/java/sootup/java/bytecode/frontend/inputlocation/RandomJarTest.java#L29C1-L33C4

palaniappan1:
> This tests probably also creates a FixJar file https://github.com/soot-oss/SootUp/blob/b6f1aac5a45be1eb0264ac400bd59fdead7809e7/sootup.java.bytecode.frontend/src/test/java/sootup/java/bytecode/frontend/inputlocation/RandomJarTest.java#L29C1-L33C4

The if (records.isEmpty()) {} block will not let write the content to the file

JonasKlauke:
what is the test doing?
testing the file writing?
the test has no assertion

palaniappan1:
> what is the test doing? testing the file writing? the test has no assertion

When running 100 jars (in a github action), if any jar fails, we know the method signature which failed, and this test writes an testcase with the jarname, method signature and the view (this test will fail usually). So making this test pass is the work for us.

JonasKlauke:
shouldnt this whole test run in one test method?

palaniappan1:
> shouldnt this whole test run in one test method?

Yes in general it should be, but here i am trying to invoke this writetest from a github action. For some reason, i cant invoke this as a general method (it needs to be annotated with test, in order to work). Yes I completely agree that if the code is invoked from general java test/java file i would have it as a helper method, but here it was not possible (from my experiece with github actions)

JonasKlauke (approved):

PR #1333, opened 2025-06-25
Add support for var-args method calls with @PolymorphicSignature causing unmatched method signatures

### LINKED ISSUE #78: Create Backport for MethodHandle/VarHandle
We should create a backport for the MethodHandle/VarHandle feature from Soot into SootReloaded https://github.com/Sable/soot/pull/1109

### LINKED ISSUE #1266: [Feature]: Investigate varargs parameter in the call graph algorithm
### Missing Feature

The varargs parameter might not be correctly handled in the call graph algorithm.
example:
void m (Object... args)

moebarni:
issue: #1266


sootup.callgraph/src/test/java/sootup/callgraph/CallGraphTestBase.java
```diff
@@ -41,7 +42,7 @@ public abstract class CallGraphTestBase<T extends AbstractCallGraphAlgorithm> {
 
   private JavaView createViewForClassPath(String classPath) {
     List<AnalysisInputLocation> inputLocations = new ArrayList<>();
-    inputLocations.add(new DefaultRuntimeAnalysisInputLocation());
+    inputLocations.add(new DefaultRuntimeAnalysisInputLocation()); // SourceType.Application
```
moebarni: unnecessary comment -> delete

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -61,10 +63,42 @@ public abstract class AbstractCallGraphAlgorithm implements CallGraphAlgorithm {
...
+        .flatMap(javaSootClass -> javaSootClass.getMethods().stream())
+        .forEach(
+            javaSootMethod -> {
+              Iterable<AnnotationUsage> annotationUsages = javaSootMethod.getAnnotations();
+              if (annotationUsages.equals(Collections.emptyList())) {
+                return;
+              }
+              for (AnnotationUsage annotationUsage : annotationUsages) {
+                if (annotationUsage.getAnnotation().equals(polymorphicAnnotationType)) {
+                  polymorphicMethodSigs.add(javaSootMethod.getSignature());
```
JonasKlauke: add here a break to save runtime

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -61,10 +63,42 @@ public abstract class AbstractCallGraphAlgorithm implements CallGraphAlgorithm {
...
+    view.getClasses()
+        .filter(sootClass -> sootClass instanceof JavaSootClass)
+        .map(sootClass -> (JavaSootClass) sootClass)
+        .flatMap(javaSootClass -> javaSootClass.getMethods().stream())
+        .forEach(
+            javaSootMethod -> {
+              Iterable<AnnotationUsage> annotationUsages = javaSootMethod.getAnnotations();
+              if (annotationUsages.equals(Collections.emptyList())) {
+                return;
+              }
```
JonasKlauke: you dont need to do this. if it is empty the for loop is just not executed once

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -667,4 +690,30 @@ protected static Optional<? extends SootMethod> findDefaultMethod(
...
+
+  /**
+   * Attempts to resolve a previously unresolved method signature by comparing it with all known
+   * method signatures annotated with {@code @PolymorphicSignature}. If a match is found (ignoring
+   * the parameters), the corresponding SootMethod is returned.
+   *
+   * @param targetMethodSignature the signature of the searched method
+   * @return the found method object, or null if the method was not found.
+   */
+  protected Optional<? extends SootMethod> findMatchingPolymorphicMethod(
```
JonasKlauke: mmh the name is missleading maybe findMatchingVarArgsMethod

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -667,4 +690,30 @@ protected static Optional<? extends SootMethod> findDefaultMethod(
...
+      @NonNull MethodSignature targetMethodSignature) {
+    for (MethodSignature polymorphicMethodSigs : preanalysis) {
+      if (targetMethodSignature.getDeclClassType().equals(polymorphicMethodSigs.getDeclClassType())
+          && targetMethodSignature.getName().equals(polymorphicMethodSigs.getName())) {
+        // return the actualTargetMethodOpt
+        return view.getMethod(polymorphicMethodSigs).map(sootMethod -> (SootMethod) sootMethod);
```
JonasKlauke: put SootMethod directly in the list to save the lookup

sootup.callgraph/src/main/java/sootup/callgraph/RapidTypeAnalysisAlgorithm.java
```diff
@@ -96,13 +97,21 @@ protected List<ClassType> collectInstantiatedClassesInMethod(SootMethod method)
...
+                value -> {
+                  // ClassType as returnType
+                  if (value instanceof AbstractInvokeExpr && value.getType() instanceof ClassType) {
+                    return Stream.of((ClassType) value.getType());
+                    // new-expression
+                  } else if (value instanceof JNewExpr) {
+                    return Stream.of(((JNewExpr) value).getType());
+                  }
+                  return Stream.empty();
+                })
```
JonasKlauke: we are only checking for new expressions. Method calls are not considered because first you do not check if they are constructor and second because if you would check for constructors you would consider supertypes as instaniated if the constructor contains a super() call.

sootup.callgraph/src/main/java/sootup/callgraph/RapidTypeAnalysisAlgorithm.java
```diff
@@ -317,8 +330,9 @@ protected void preProcessingMethod(
     if (method == null) {
       return;
     }
-
+    System.out.println("PreProcessed Method: " + method);
```
JonasKlauke: no sout in the code

sootup.callgraph/src/main/java/sootup/callgraph/RapidTypeAnalysisAlgorithm.java
```diff
@@ -317,8 +330,9 @@ protected void preProcessingMethod(
...
     List<ClassType> newInstantiatedClasses = collectInstantiatedClassesInMethod(method);
+    System.out.println("New Instantiated Classes: " + newInstantiatedClasses);
```
JonasKlauke: same here

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -667,4 +690,30 @@ protected static Optional<? extends SootMethod> findDefaultMethod(
...
+   * the parameters), the corresponding SootMethod is returned.
+   *
+   * @param targetMethodSignature the signature of the searched method
+   * @return the found method object, or null if the method was not found.
+   */
+  protected Optional<? extends SootMethod> findMatchingPolymorphicMethod(
+      @NonNull MethodSignature targetMethodSignature) {
+    for (MethodSignature polymorphicMethodSigs : preanalysis) {
+      if (targetMethodSignature.getDeclClassType().equals(polymorphicMethodSigs.getDeclClassType())
+          && targetMethodSignature.getName().equals(polymorphicMethodSigs.getName())) {
```
JonasKlauke: two cases are not covered which me might have to investigate.
1. different parameters,
2. subclasses of classes that contain these methods

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -667,4 +690,30 @@ protected static Optional<? extends SootMethod> findDefaultMethod(
...
+   * Attempts to resolve a previously unresolved method signature by comparing it with all known
+   * method signatures annotated with {@code @PolymorphicSignature}. If a match is found (ignoring
+   * the parameters), the corresponding SootMethod is returned.
+   *
+   * @param targetMethodSignature the signature of the searched method
+   * @return the found method object, or null if the method was not found.
+   */
+  protected Optional<? extends SootMethod> findMatchingPolymorphicMethod(
+      @NonNull MethodSignature targetMethodSignature) {
+    for (MethodSignature polymorphicMethodSigs : preanalysis) {
```
JonasKlauke: we need a fast checkup that checks if it is even possible to find a varargs target.
I think we have to rethink the datastructure. we should have fast checkups with the methodname

sootup.callgraph/src/test/resources/callgraph/Polymorphic/source/MethodHandleInvokeExample1.java
```diff
@@ -0,0 +1,14 @@
...
+import java.lang.invoke.MethodType;
+
+public class MethodHandleInvokeExample1 {
+    public static void main(String[] args) throws Throwable {
+        MethodHandle handle = MethodHandles.lookup()
+                .findVirtual(String.class, "length", MethodType.methodType(int.class));
+
+        handle.invoke("Hello");
+    }
+}
```
JonasKlauke: mmh i think we can delete this test. I like Example 2 and I think both should test the same thing

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -61,10 +63,42 @@ public abstract class AbstractCallGraphAlgorithm implements CallGraphAlgorithm {
 
   @NonNull protected final View view;
   @NonNull protected final TypeHierarchy typeHierarchy;
+  @NonNull protected final List<MethodSignature> preanalysis;
```
JonasKlauke: use guavas table
this allows to map SootMethod to the method name and the class name.
If we need to differ between parameters put in a list as value type
https://www.baeldung.com/guava-table
the table also allows to get all fitting elements for a method name. if we need to find a class by using the super classes

This should allows us to have fast checkups

JonasKlauke (changes requested):

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
moebarni: Created tests.

moebarni: No subclasses of java.lang.invoke.MethodHandle/VarHandle possible, both classes have package-private constructors and all their implementations are package-private.

JonasKlauke: can you access the annotation outside of the package?

moebarni: No: 'java.lang.invoke.MethodHandle.PolymorphicSignature' is not public in 'java.lang.invoke.MethodHandle'. Cannot be accessed from outside package

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
moebarni: Changed data structure.

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
moebarni: new first check: contains the table the targetName? via table.containsRow(name)

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
moebarni: JavaSootMethod directly saved in table; later Optional.of returned

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
moebarni: Changed to directly invoking table.get(methodName, classTypeName) resulting in the corresponding VarArgsSootMethod or null; returning the Optional of the SootMethod or an empty Optional (if result is null)

JonasKlauke:
fixes also #78

sootup.callgraph/src/main/java/sootup/callgraph/RapidTypeAnalysisAlgorithm.java
moebarni: Rollback applied. New approach required to treat java.lang.invoke.MethodHandle (with its package-private constructor) as instantiated during RTA analysis.

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -61,10 +66,59 @@ public abstract class AbstractCallGraphAlgorithm implements CallGraphAlgorithm {
...
+                if (annotationUsage.getAnnotation().equals(polymorphicAnnotationType)) {
+                  ClassType classType = javaSootMethod.getDeclaringClassType();
+                  polymorphicTypeAndMethods
+                      .computeIfAbsent(classType, k -> new ArrayList<>())
+                      .add(javaSootMethod);
+                  polymorphicMethods.put(
+                      javaSootMethod.getName(), classType.getFullyQualifiedName(), javaSootMethod);
+                }
+              }
+            });
```
JonasKlauke: transfrom this to a filter

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -61,10 +66,59 @@ public abstract class AbstractCallGraphAlgorithm implements CallGraphAlgorithm {
...
+   * java.lang.invoke.MethodHandle$PolymorphicSignature}.
+   *
+   * @return a list of method signatures annotated with {@code @PolymorphicSignature}
+   */
+  protected Table<String, String, SootMethod> getMethodsWithPolymorphicAnnotation() {
+    Table<String, String, SootMethod> polymorphicMethods = HashBasedTable.create();
+    ClassType polymorphicAnnotationType =
+        view.getIdentifierFactory()
+            .getClassType("java.lang.invoke.MethodHandle$PolymorphicSignature");
+    Map<ClassType, List<JavaSootMethod>> polymorphicTypeAndMethods = new HashMap<>();
```
JonasKlauke: we dont need to save this.
Small Info: In this case a ArrayListMultimap would be easier to manage instead of a Map containing a List

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -61,10 +66,59 @@ public abstract class AbstractCallGraphAlgorithm implements CallGraphAlgorithm {
...
+      typeHierarchy
+          .subtypesOf(classType)
+          .forEach(
+              subtype -> {
+                for (JavaSootMethod javaSootMethod : javaSootMethods) {
+                  polymorphicMethods.put(
+                      javaSootMethod.getName(), classType.getFullyQualifiedName(), javaSootMethod);
+                }
+              });
+    }
```
JonasKlauke: do this in a foreach or in a flatmap + foreach

JonasKlauke (changes requested):

PR #1349, opened 2025-07-24
Convert stmts to PAG edges

kadirayk:


sootup.spark/src/main/java/sootup/spark/PAG.java
```diff
@@ -0,0 +1,45 @@
...
+      addEdge(source, target, PAGEdge.allocation());
+    } else {
+      log.error("Invalid edge type for source: {} -> target: {}", source, target);
+    }
+  }
+
+  private void addEdge(Node source, Node target, PAGEdge edge) {
+    getDelegate().addVertex(source);
+    getDelegate().addVertex(target);
+    getDelegate().addEdge(source, target, edge);
```
JonasKlauke: should we check if they already exist?

sootup.spark/src/main/java/sootup/spark/PAG.java
```diff
@@ -0,0 +1,45 @@
...
+import sootup.spark.node.Node;
+import sootup.spark.node.VariableNode;
+
+/** Pointer assigment graph */
+@Slf4j
+@Getter
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+public class PAG {
+
+  @NonNull Graph<Node, DefaultEdge> delegate = new DefaultDirectedGraph<>(DefaultEdge.class);
```
JonasKlauke: In Call Graph we have a Call Class which is given as type of call. 
It contains the source, target and stmt.
We could do here a similar approach. or it should at least use PAGEdge

sootup.spark/src/main/java/sootup/spark/PAG.java
swissiety: lets cache the getDelegate() calls

JonasKlauke: we can simply use the field

sootup.spark/src/main/java/sootup/spark/PAG.java
kadirayk: These are not call edges, rather assignment edges, e.g., between Locals.

JonasKlauke (approved):

PR #1361, opened 2025-08-06
split OverridingJavaClassSource into two subclasses

SamarthBengle:


sootup.java.core/src/main/java/sootup/java/core/StandaloneJavaClassSource.java
```diff
@@ -0,0 +1,502 @@
...
+import sootup.core.model.SootField;
+import sootup.core.model.SootMethod;
+import sootup.core.types.ClassType;
+import sootup.core.util.CollectionUtils;
+import sootup.java.core.types.JavaClassType;
+
+@SuppressWarnings({"unchecked"})
+public class StandaloneJavaClassSource extends OverridingJavaClassSource {
+
+    private static final Position NO_POSITION =
```
swissiety, line 20: an equivalent construct already exists in..Position?

swissiety (commented):
please clean up the old/existing class - is that deletedMethods etc. newly introduced logic?

SamarthBengle:
> please clean up the old/existing class - is that deletedMethods etc. newly introduced logic?
Yes, this is new. I added it to avoid copying entire collections when deleting, it just marks what to hide instead

sootup.java.core/src/main/java/sootup/java/core/StandaloneJavaClassSource.java
SamarthBengle: I didn't find anything suitable for it. So I used like this

swissiety:
> > please clean up the old/existing class - is that deletedMethods etc. newly introduced logic?
> > Yes, this is new. I added it to avoid copying entire collections when deleting, it just marks what to hide instead

please don't - I see the thought of doing it on-demand as well but we have to recalculate the collection(s) on every call while the collections are expected to be "small" so it just adds more logic

PR #1369, opened 2025-08-21
taint analysis example

SamarthBengle:


swissiety (commented):
ok - please add a step by step guide "tutorial" written in markdown.
- [ ] write tutorial
- [ ] add a new file in docs/ .
- [ ] To include the code example please make use of a plugin that lets you include files so that you can refer to specific lines/code range in the TaintAnalysisTest.java file so that the mentioned lines are included in the markdown rendering (find/configure a mkdocs plugin if necessary) - maybe a mkdocs plugin that lets you specify code regions by a placeholder in a code comment which we could insert into the original source would be really great so the ranges would not changes if someone adds anew line of code.

docs/taint-analysis-example.md
```diff
@@ -0,0 +1,715 @@
...
+- **Entry method**: Starting point for analysis
+
+## Step 3: Implementing the Taint Analysis Problem
+
+### 3.1 Create the IFDS Problem Class
+
+The core analysis logic extends `DefaultJimpleIFDSTabulationProblem`:
+
+```java
+static class TaintAnalysisProblem extends DefaultJimpleIFDSTabulationProblem<Object, InterproceduralCFG<Stmt, SootMethod>> {
```
swissiety: can we break that huge section down into multiple smaller ones + explaination?

swissiety (changes requested):
lets make use of a mkdocs plugin to import the sections from a .java file

swissiety:
integrated into #1629

PR #1384, opened 2025-09-22
Fix: Add missing arguments for staticinvoke after DEX conversion

main-force:


swissiety:
thanks for your contribution - superseded by #1578 (has a test case)

PR #1393, opened 2025-09-24
adapt dependabot.yml

swissiety:


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    # Only run for PRs created by Dependabot - extended verification is done in the reusable workflow
+    if: github.actor == 'dependabot[bot]'
+    # These permissions are needed to approve pull requests
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Auto approve Dependabot PR
+        uses: secure-software-engineering/actions/dependabot@develop
```
JonasKlauke, line 23: that will probably cause an error like the other flows which are currently broken

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          token: ${{ secrets.AUTO_MERGE_PAT }}
```
JonasKlauke, line 25: secrets also dont work here
https://github.com/soot-oss/SootUp/actions/runs/18031767720/workflow

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  ApproveAndMerge:
+    name: Auto approve Dependabot PRs
+    runs-on: ubuntu-latest
+    # Only run for PRs created by Dependabot - extended verification is done in the reusable workflow
+    if: github.actor == 'dependabot[bot]'
```
JonasKlauke, line 16: can this be missused?

JonasKlauke (changes requested):

.github/workflows/dependabot.yml
JonasKlauke: my bad that seems to be something else

.github/workflows/dependabot.yml
swissiety: you can just configure git to use a emailaddress for commits etc - here the one of dependabot - but the gpg signature verification would fail - this is handled in the called workflow as the comment states

.github/workflows/dependabot.yml
swissiety: the path changed with those reusable actions. I rework that in another PR

JonasKlauke (approved):

PR #1435, opened 2025-10-15
sorting method calls

### LINKED ISSUE #1086: [Feature]: Method calls sorted by sequence
### Missing Feature

# callFrom(methodSignature) returns a no-sorted set, consider making a sorted list by the call sequence?
I'm currently doing a research of type-bug

### Suppose i have a program below
```java
public class A {
    private int m() { System.out.println("A.m"); return 0; }
    public static void main(String... args) { type_bug.refined.B b = new type_bug.refined.C(); b.m(); }
}

abstract class B extends A implements I {
}

class C extends B {
    public void m() { System.out.println("C.m"); }
}

interface I {
    public void m();
}
```

### i use the following code to get the Methods called in A.main()
```java
final ViewTypeHierarchy typeHierarchy = new ViewTypeHierarchy(view);
CallGraphAlgorithm cha = new ClassHierarchyAnalysisAlgorithm(view);
// Create CG by initializing CHA with entry method(s)CallGraph cg = cha.initialize(Collections.singletonList(methodSignature));
cg.callsFrom(methodSignature).forEach(System.out::println);
```

i got the following output:
<type_bug.refined.C: void m()>
<type_bug.refined.C: void <init>()>
<type_bug.refined.I: void m()>

while the correct sequence should be like:
<type_bug.refined.C: void <init>()>
<type_bug.refined.C: void m()>
<type_bug.refined.I: void m()>

i wonder if there is a way to sort the method call correctly😊

ashik-18:
Closes [#1086](https://github.com/soot-oss/SootUp/issues/1086)

### DIFF
--- a/sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
+++ b/sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
@@ -22,16 +22,12 @@ package sootup.callgraph;
  * #L%
  */
 
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 import org.jgrapht.graph.DirectedPseudograph;
 import org.jspecify.annotations.NonNull;
 import sootup.core.jimple.common.stmt.InvokableStmt;
+import sootup.core.model.LinePosition;
 import sootup.core.signatures.MethodSignature;
 import sootup.core.signatures.SootClassMemberSignature;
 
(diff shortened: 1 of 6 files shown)


sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
```diff
@@ -139,7 +135,37 @@ public Set<MethodSignature> callSourcesTo(@NonNull MethodSignature targetMethod)
...
+    var pos = stmtPositionOf(c);
+    return pos instanceof LinePosition lp ? lp.getFirstCol() : Integer.MAX_VALUE;
+  }
+
+  private static int getColEndFrom(Call c) {
+    var pos = stmtPositionOf(c);
+    return pos instanceof LinePosition lp ? lp.getLastCol() : Integer.MAX_VALUE;
+  }
+
+  private static Object stmtPositionOf(Call c) {
```
swissiety: more precise return type. please move logic into a custom Comparator class

swissiety (changes requested):
would be great if we can make the sorting optional / another method / ...

sootup.callgraph/src/main/java/sootup/callgraph/GraphBasedCallGraph.java
ashik-18: I added a new method to sort the calls and also had to implement it in OnFlyCallGraph. Some existing methods in OnFlyCallGraph throw UnsupportedOperationException, so I wasn’t sure if sorting is needed, let me know if it should be removed.

swissiety (approved):

PR #1437, opened 2025-10-17
Fix DOI format in CITATION.cff

jlinx:
The DOI seems to be wrong. This fixes it...


swissiety (approved):
:+1: thanks for finding this typo

PR #1459, opened 2025-11-26
[Feature]: Default constructor for CpgCreator

rashmisoni2003:
```
This PR adds a no-argument (default) constructor to the CpgCreator class so it can be instantiated without explicitly providing AstCreator, CfgCreator, CdgCreator, and DdgCreator instances.
```


JonasKlauke:
can you run fmt:format or simply mvn install to fix the style and commit the changes

stschott (approved):

PR #1522, opened 2026-02-16
Add additional Callgraph class inclusion filtering

aburghuber:
## What does this PR do?

Adds a nullable parameter called `basePackageName` to `CallGraphAlgorithm#initialize` that filters out all non-source classes from the graph (more accurately). Also adds `AbstractCallGraphAlgorithm#includeCallToTarget` for even more additional filtering.

The motivation behind this is that in Android projects often Android base API classes were included in the graph.

---

### Code style & guidelines
- [X] I ran the formatter: `mvn com.spotify.fmt:fmt-maven-plugin:format`
- [X] I added the necessary comments in the code
- [X] I provided meaningful tests for my proposed change - Not needed
- [X] I updated documentation (if needed)

### Self-review
- [X] I performed a self-review of my code
- [X] I added or updated tests where needed
- [X] My branch is up to date with `develop`

### Review
- [ ] CI checks are green

### DIFF
--- a/sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
+++ b/sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
@@ -104,10 +104,24 @@ public abstract class AbstractCallGraphAlgorithm implements CallGraphAlgorithm {
    * @return the complete constructed call graph starting from the entry methods.
    */
   @NonNull
+  final CallGraph constructCompleteCallGraph(List<MethodSignature> entryPoints) {
+    return constructCompleteCallGraph(entryPoints, null);
+  }
+
+  /**
+   * This method starts the construction of the call graph algorithm. It initializes the needed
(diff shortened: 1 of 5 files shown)


JonasKlauke:
mmh I like the includeCallTarget method. But I think a Call object as parameter or the addition of the invoking statement would be better, so the users could also consider the stmt in the check.

i dont like the part about the base package. An application might not have one base package. We have also a similar approach with tagging a inputlocaction as Library. If it is tagged as Library then the call graph ignores it.
with a combination of your includeCallTarget method it should be possible to create your base package solution as an user.

So i would only partly accept this PR. However, thanks for improving SootUp

ericbodden:
@swissiety could you check this out please? It has been open for a while...

swissiety:
Thanks @aburghuber for your contribution! If I'm understanding your implementation correctly, I already have a more general/extensible approach in my local backlog that will cover the same use case and therefore will supersede this PR (#1665). If I'm mistaken please feel free to reopen this PR. Thanks for taking the initiative to improve SootUp!

PR #1540, opened 2026-03-10
Minor enhancement to copy propagator

SanthoshMohan97:
Evaluating some trivial binary expressions to help propagate further values

## What does this PR do?
Consider this code fragment 

        1. int a = 3 * 10 + 7;
        2. int b = a + 5;
        3. int c = a + b

a's value could be determined by java compiler, during copy propagation
              1. a = 37; 
              2.  b = 37 + 5; (uncomputed expression) and
              3. c = 37 + b; (b still stays b because its not a constant)

After the changes. binary operations in 2 would be computed and simplfied.
           1.  a = 37;
           2. b = 42;
           3. c = 37 + 42 ==> c = 79;

Note: This could NOT be acheived by pipelining of copy propagation and constantfolding interceptors also.

Changed the corresponding test case for copy propgator


swissiety:
you could try to reuse the Evaluator to support some more cases.

swissiety (approved):
👍

PR #1544, opened 2026-03-15
Add getValue method to BooleanConstant class

ericbodden:
Currently there was no way to extract BooleanConstant's value. I added an accessor.


swissiety (approved):
:+1:

PR #1596, opened 2026-05-20
Jimple string analysis inputlocation fix

Hazem-Gamall:
## What does this PR do?
return an empty optional when the class is not found in the Jimple string

---

## Linked issue (if any)
- Fixes #<issue-number>
  

---

### Code style & guidelines

### Self-review

### Review
- [ ] CI checks are green


sootup.jimple.frontend/src/main/java/sootup/jimple/frontend/JimpleStringAnalysisInputLocation.java
```diff
@@ -85,6 +85,14 @@ private OverridingJavaClassSource getOverridingClassSource(
   @Override
   public Optional<? extends SootClassSource> getClassSource(
       @NonNull ClassType type, @NonNull View view) {
+    OverridingJavaClassSource classSource =
+        getOverridingClassSource(jimpleFileContents, bodyInterceptors, view);
```
swissiety: lets cache the results of getOverridingClassSource for line 96

swissiety (changes requested):

swissiety (approved):

swissiety (changes requested):
lets either store the raw string or the parsed classSource

Hazem-Gamall:
> lets either store the raw string or the parsed classSource

Both are called in very different contexts, the Jimple string is necessary and is called by the user of SootUp to set-up the string to be parsed later, then the view itself calls upon the getClassSource() method to parse and return the file.
So the string can't be parsed at `getClassSource()` since the user knows it but not the view, and the class can't be parsed in the constructor since the view is not there yet.
Also parsing the string on every call is wasteful, I guess for performance reasons we can null out the string after first parse, but again what if a different view is passed, the class allows it, and then we'd have to parse the method again.
So what do you suggest?

swissiety (approved):

PR #1653, opened 2026-07-13
Feature/bytecode type use annotations

kadirayk:

The bytecode (ASM) frontend dropped JSR 308 `TYPE_USE` type annotations. Return-type (`@Nat int foo()`), formal-parameter (`void foo(@Nat int n)`), and local-variable (`@Nat int x`) type annotations were not read. It also ignored `RuntimeVisibleParameterAnnotations` (only invisible ones were read).

Additive changes in `AsmMethodSource`/`AsmClassSource`:
- Return-type & parameter `TYPE_USE`: filter `visible/invisibleTypeAnnotations` by `TypeReference` sort (`METHOD_RETURN` / `METHOD_FORMAL_PARAMETER`+index). Return-type ones are merged into the method's annotations; parameter ones are attached to the parameter `Local`. Also now reads visible parameter declaration annotations.
- Local-variable `TYPE_USE`: read `visible/invisibleLocalVariableAnnotations`, matched by slot index + scope range (with a slot-only fallback, since a local is created at its defining store just before its declared scope), attached via a new `createUniqueLocal(name, type, annotations)`.

Limitation (noted in-code): `typePath` (position within nested/generic types) isn't represented by `AnnotationUsage`. Tests: `TypeUseAnnotationTest` + source subject + compiled `.class` fixtures, covering return/parameter/local in `RUNTIME` and `CLASS` retention.

**Checklist**

*Code style & guidelines*

*Self-review*

*Review*
- [ ] CI checks are green
- [x] I requested a review from a core contributor


sootup.java.bytecode.frontend/src/main/java/sootup/java/bytecode/frontend/conversion/AsmClassSource.java
```diff
@@ -120,16 +120,24 @@ public Set<JavaSootMethod> resolveMethods() throws ResolveException {
...
 
+              // Method annotations: declaration annotations (RuntimeVisible/InvisibleAnnotations,
+              // e.g. @Pure) plus JSR 308 return-type annotations (TYPE_USE on the return type,
+              // e.g. @Nat int foo()). Parameter type annotations are attached to the parameter
+              // Locals instead (see AsmMethodSource#buildPreambleLocals).
+              List<AnnotationUsage> methodAnnotations =
+                  Streams.concat(
+                          convertAnnotation(methodSource.visibleAnnotations),
+                          convertAnnotation(methodSource.invisibleAnnotations))
+                      .collect(Collectors.toList());
```
swissiety: lets use an in/out parameter for collecting the result here as well, so we can replace the expensive stream object intitialization etc.

sootup.java.bytecode.frontend/src/main/java/sootup/java/bytecode/frontend/conversion/AsmMethodSource.java
```diff
@@ -336,14 +338,70 @@ private String determineLocalName(int idx, @Nullable AbstractInsnNode atInsn) {
...
+        if (insnIdx >= startIdx && insnIdx < endIdx) {
+          rangeMatch = true;
+          break;
+        }
+      }
+      if (rangeMatch || slotMatch) {
+        // Note: node.typePath (nested-type position) is not represented by AnnotationUsage.
+        out.add(AsmUtil.createAnnotationUsage(node));
+      }
+    }
```
swissiety: time complexity seems bad on first glance as we do the loop for every Local: lets refactor that so that we collect the information in a reasonable datastructure for all the locals at once and reuse that result/datastructure?

swissiety (changes requested):
Hi external contributor @kadirayk :)

swissiety (approved):
👍

PR #1657, opened 2026-07-14
feat(qilin-pta): update+rework integration

swissiety:
The qilin module was in a bad state. global singleton, no concurrency, not integrated into the rest of the pointer/callgraph framework. a seperate module somehow using sootup (from a blueprint how to use soot).

- [x] refactored static modifiers away: enables concurrent pointer analyses/ callgraph generation  
- [x] replaced string config with enums/constants/builder
- [x] opening up previously hardcoded: native methods / lambda handling
- [x] remove qilin cli direct access that was used in the qilin paper to benchmark it: not needed in this library, can be used/implemented/wired from another project if desired.
- [x] adapt the tests which currently fail silently as artifacts folder (~1G)  is currently not tracked in git
- [x] address PTAUtils.resetMethodBodyCache()`
- [ ] make use of Signatures instead of Strings for commonly used Defs (JavaDefinitions) 
- [x] unify interface with spark?
- [x] integrate CallResolver (#1665)
- [x] integrate latest fixes&improvements from https://github.com/QilinPTA/Qilin/commits/main/
- [x]  move Tamiflex reflection-log parser  to sootup.core (shared by callgraph+qilin)
- [ ] parallelize


sootup.qilin/src/main/java/qilin/core/pag/PAG.java
```diff
@@ -103,6 +113,7 @@ public class PAG {
 
   public PAG(PTA pta) {
     this.pta = pta;
+    PTAUtils.resetMethodBodyCache();
```
JonasKlauke: mmh this looks like something that effects other PAG that are maybe running in the background.

sootup.qilin/src/test/java/qilin/test/core/InvokeDynamicTests.java
```diff
@@ -0,0 +1,120 @@
...
+        new File(
+            rootDir,
+            "artifact"
+                + File.separator
+                + "benchmarks"
+                + File.separator
+                + "JREs"
+                + File.separator
+                + "jre1.8.0_121_debug");
+    jre8Path = jreFile.getCanonicalPath();
```
JonasKlauke: can we replease this with the default input location?

JonasKlauke (approved):

sootup.qilin/src/test/java/qilin/test/core/InvokeDynamicTests.java
swissiety: especially as the whole artifacts folder is currently not tracked in git and the tests fail silently -> WIP ;)

PR #1665, opened 2026-07-16
CallGraph Scope

swissiety:
filters calls to be explored.


JonasKlauke (changes requested):
I feel like if we do this with a scope we should do something similar with the edge filter.

```java
  /**
   * Decide whether a call from <code>method</code> represented by <code>statement</code> shall be
   * added to the call graph. Default: accept everything. Subclasses can override this method to
   * implement pruning.
   *
   * @param method the source (caller) method
   * @param statement the invokable statement causing the call
   * @return true if the call should be included in the call graph
   */
  protected boolean includeCall(@NonNull SootMethod method, @NonNull InvokableStmt statement) {
    return true;
  }
```

otherwise it is looking good

JonasKlauke (approved):

Issue #1684, opened 2026-07-23
[Feature]: Decide whether MutableTypeHierarchy#addType should support classes its own View can't resolve


Hazem-Gamall:
### Missing Feature

`MutableTypeHierarchy#addType(SootClass)` currently accepts any `SootClass` instance, including one the hierarchy's own backing `View` cannot independently resolve via `getClass(...)` - e.g. a class built by hand, or resolved through a completely different `View`. This isn't hypothetical: the existing `ViewTypeHierarchyTest#addType` test (predates any of the changes below) already constructs a `SootClass` this way and adds it successfully - it's part of the type's established, tested public contract.

The problem is that `TypeHierarchy#subtypeClassesOf` (and, more generally, any code in `ViewTypeHierarchy` that needs to turn a `ClassType` already in the hierarchy graph into an actual `SootClass`) does this by calling `view.getClass(type)`. For a type added this way, that call returns empty, so the type is silently excluded from `subtypeClassesOf`'s results - even though it's correctly present in the graph-only `subtypesOf`. There's currently no way to ask the hierarchy "give me the SootClass for this type" and reliably get one back for every type it knows about.

This matters in practice for pipelines that discover classes through a separate mechanism from SootUp's own `View` resolution (e.g. combining SootUp with another analysis tool that resolves its own classes and wants to register them into SootUp's type hierarchy) and then expect full hierarchy queries - not just graph/`ClassType`-level ones - to see them.

Worth deciding one way or the other:
- Formally support it: give `MutableTypeHierarchy` (or the `View`) a way to resolve a `SootClass` for such types without going through `view.getClass()`, and update every place in `ViewTypeHierarchy` that does the "resolve `ClassType` from the graph, then `view.getClass()` it" pattern to use that instead.
- Or formally reject/restrict it: document that `addType` requires a `View`-resolvable class and consider validating that at the API boundary, so the current silent-exclusion behavior isn't a surprise.

Currently `subtypeClassesOf` documents and tests the silent-exclusion behavior as intentional (see `ViewTypeHierarchy`/`TypeHierarchy` javadoc and `ViewTypeHierarchyTest#addTypeInvalidatesSubtypesOfCacheEvenForClassesUnresolvableByView`), but that was a scoping decision for a specific performance PR, not a considered answer to this question.

Hazem-Gamall:
Make it not publicly accessible.

PR #1685, opened 2026-07-23
Cache expensive, repeatedly-queried type hierarchy and CHA dispatch resolution

Hazem-Gamall:
CHA callgraph construction is extremely slow on real-world code. Profiling a real ~10MB APK analysis (FlowDroid callgraph construction followed by SootUp CHA construction over the combined method set, via JFR) traced this to several operations that, while individually correct, redid expensive work on every single call instead of once: `SootClass.getMethod(subSignature)` linearly scanned every member; `MethodSubSignature`/`SootClassMemberSubSignature.hashCode()` was recomputed on every access despite being used heavily as Map/Set keys during hierarchy and dispatch resolution; `ViewTypeHierarchy.subtypesOf`/`subtypeClassesOf` re-walked the class graph on every query, even repeated queries for the same type; and CHA's virtual/interface dispatch resolution recomputed the full subtype/target-method search on every call, even when it had already resolved that exact target signature before.

- `JavaSootClass.getMethod(subSignature)`: build (and lazily cache) a `Map<MethodSubSignature, SootMethod>` once per class instead of a linear scan on every lookup.
- `MethodSubSignature` / `SootClassMemberSubSignature`: memoize `hashCode()` via `Suppliers.memoize`, matching the existing `_cachedToString` pattern already used elsewhere in the codebase.
- `ViewTypeHierarchy`: cache `subtypesOf`/`subtypeClassesOf` results per queried type, invalidated wholesale on `addType()` (the hierarchy's only mutator). Adds `TypeHierarchy.getModificationCount()` - a simple monotonically increasing counter - so external caches derived from hierarchy queries (see below) can detect staleness without a push/listener model. `subtypeClassesOf` resolves each subtype to a `SootClass` via the underlying `View`'s own `getClass()`; a type added via `addType()` with a `SootClass` the view can't independently resolve is (intentionally) excluded from its result, same as it always was.
- `ClassHierarchyAnalysisAlgorithm`: cache virtual/interface dispatch resolution keyed by target `MethodSignature`, invalidated via the hierarchy's modification counter. Verified this only pays the resolution cost once per unique target signature - even when queried from many different call sites - via a dedicated test that spies on the `TypeHierarchy` with Mockito to deterministically assert the underlying hierarchy call count, plus a manual benchmark against the real JDK runtime jar (see numbers below). `RapidTypeAnalysisAlgorithm`'s own dispatch resolution isn't safely cacheable the same way (its `instantiatedClasses` set mutates during the run), so it instead just reuses the now-cached `TypeHierarchy.subtypeClassesOf` directly - a one-line change.

Real before/after numbers, measured directly against `java.base` (`DefaultRuntimeAnalysisInputLocation`), resolving `Object.toString()` (2403 dispatch targets) via direct `resolveCall()` calls to isolate repeated resolution of the same target from the one-time cost of scanning the JDK's class structure (~9.9s either way, unaffected by this change):

| | Before | After |
|---|---|---|
| Each repeated resolution of the same virtual dispatch target | ~165.7ms | ~0.006ms |

On the motivating real-world case (a ~10MB APK analyzed via FlowDroid + SootUp), CHA callgraph construction time dropped from ~58 minutes to single-digit minutes across this and the graceful-resolution fix in the stacked PR below.

**Real-world profiling results (CHA & RTA, JFR, same ~10MB APK)**

To get a clean before/after on real application code, both call graph algorithms were profiled twice each - once on this PR's parent commit (root-cause-1 fix only, no caching from this PR) and once with this PR applied - using JFR, on the same APK/analysis. All numbers below exclude the shared ~4-minute FlowDroid callgraph construction prefix (fixed cost, identical in every run, unaffected by this PR) to isolate SootUp's own contribution.

*Wall-clock time (SootUp portion only):*

| | Before | After | Speedup |
|---|---|---|---|
| CHA | 2756 s (~45.9 min) | 454 s (~7.6 min) | **~6.1x** |
| RTA | 2585 s (~43.1 min) | 481 s (~8.0 min) | **~5.4x** |

*GC-active window (span between first and last GC event):*

| | Before | After | Shrinks by |
|---|---|---|---|
| CHA | 1495.6 s | 345.7 s | **~4.3x** |
| RTA | 1411.7 s | 358.6 s | **~3.9x** |

*Heap occupancy while GC is active - essentially unchanged before/after:*

| | CHA before | CHA after | RTA before | RTA after |
|---|---|---|---|---|
| Peak heap used | 3788.8 MB | 3891.2 MB | 3788.8 MB | 3788.8 MB |
| Avg heap used | 2123.9 MB | 2172.8 MB | 2211.0 MB | 2073.6 MB |

This is the key takeaway on the memory side: the fix doesn't reduce how much heap is needed at any given moment - it reduces how long the JVM spends generating garbage in the first place.

*Allocation profile:*

| | CHA before | CHA after | RTA before | RTA after |
|---|---|---|---|---|
| Total allocation samples | 341,242 | 63,069 (~5.4x fewer) | 339,944 | 63,475 (~5.4x fewer) |
| GC cycles | 751 | 273 (~2.75x fewer) | 693 | 301 (~2.3x fewer) |
| `java.util.stream.ReferencePipeline$Head` samples | 98,482 | 2,044 (**~48x fewer**) | 99,124 | 1,756 (**~56x fewer**) |

Before this PR, the single largest source of allocation by far - in both CHA and RTA - was Stream pipeline scaffolding (`ReferencePipeline$Head/2/3/7`, `AbstractPipeline$$Lambda`, `Spliterators$ArraySpliterator`, `Streams$StreamBuilderImpl`, `ArrayListSpliterator`, `FindOps$FindSink$OfRef`), generated fresh on every uncached `subtypesOf(...).flatMap(...)` composition and every linear `getMethod` scan. After this PR, that category drops out of the top allocated types entirely, replaced by the cache's own (much cheaper) `HashMap`/`ArrayList` bookkeeping. CHA and RTA independently confirm the same magnitude of improvement, as expected since both funnel through the same underlying caches.

Stacked on #1677 (graceful handling of unresolvable types) - both were developed together against the same profiled workload.

**Linked issue (if any)**
#1684

**Checklist**

*Code style & guidelines*

*Self-review*

*Review*
- [ ] CI checks are green


swissiety (approved):
👍

eortega-pjr:
We just finished porting to 3.0.1 and I believe this patch helped drop our CHA time for a particular test from 24m to 42s! Assuredly other patches helped but I'm betting this one was the most impactful.

PR #1692, opened 2026-07-30
perf: cache absent classes in JavaView.getClass

ericbodden:
## Problem

`JavaView.getClass` memoizes successful resolutions in the `ClassCache` but not failures. On a miss it re-runs `getClassSource` every time, which probes every `AnalysisInputLocation` — including the JDK jimage filesystem — while holding the view's monitor.

Call graph construction against an incomplete classpath does this constantly. Measured over a full CHA `initialize()` on SootUp's own classes, with its dependency jars deliberately absent from the input locations:

| metric | value |
| --- | --- |
| `getClass` calls | 102,808 |
| misses | 1,457 (1.4%) |
| **distinct** missed types | **59** |
| redundant re-lookups | 1,398 |

The 59 types are guava, commons-lang3, slf4j, asm and friends — the ordinary situation when analysing an application without its full dependency set on the Soot classpath.

## Fix

Remember absent types, so a repeat miss costs one hash lookup. The positive cache is still consulted first, so the hot path is unchanged.

## Results

| workload | before | after | |
| --- | --- | --- | --- |
| 455 app classes / 4,150 entry points | 285 ms | ~150 ms | **1.9×** |
| 12 app classes / 179 entry points | 102 ms | ~19 ms | **5.4×** |

The mechanism is directly visible: the same 1,457 misses over 59 distinct types still happen, but **time spent in them drops from 271 ms to 21 ms**.

The ratio grows as the analysis shrinks, since the miss cost is roughly fixed in the number of distinct absent types. With a **complete** classpath there are 8 misses (1 distinct) and no measurable change (1.03×, noise) — this fix costs nothing when it isn't needed.

**Call graphs are identical.** Verified equal method and edge counts (13,577 / 335,355) with and without the cache on identical bytecode.

## Correctness

Entries stay valid because the set of input locations is fixed for the lifetime of the view.

`MutableJavaView.addClass` additionally clears the record via a new `protected forgetAbsence(ClassType)`. To be explicit: **that call is not load-bearing today** — `getClass` consults the class cache before the absence set, and `addClass` populates that cache, so a stale record is shadowed anyway. It is kept as insurance against a reordering of those two checks, or a mutating view backed by an evicting cache such as `LRUCache`. This is stated in the javadoc so nobody mistakes it for a fix to an active bug.

## Tests

- `JavaViewTest.testAbsentClassIsResolvedOnlyOnce` — counts input-location lookups rather than timing them, so it is not sensitive to machine speed. Verified load-bearing: without the cache it fails with `expected: <1> but was: <20>`.
- `MutableSootClientTest.classAdditionAfterFailedLookupTest` — pins the invariant that a failed lookup cannot poison a later `addClass`. Labelled in its javadoc as an invariant test, not a regression test for `forgetAbsence`, because it passes with or without that call.

`sootup.java.core` (91), `sootup.callgraph` (65) and `sootup.tests` (98) all pass.

## Notes for reviewers

- **Where absence lives.** Kept in `JavaView` rather than the `ClassCache` interface, which would mean changing the interface plus `FullCache`, `LRUCache` and `MutableFullCache`, and deciding whether an LRU cache should evict absence entries. Happy to move it if you'd prefer it behind the cache abstraction.
- **Not included, needs a decision.** `inputLocations` is a `final` *reference* assigned without a defensive copy, so a caller retaining its list could mutate it post-construction. The "absence is permanent" argument therefore rests on input locations being fixed *by contract*. A `List.copyOf(...)` in the constructor would make it true by construction — and would match the class javadoc's existing claim that the view "cannot be altered after its creation" — but it's a behaviour change, so I left it out.
- **Pre-existing, unrelated.** `sootup.java.bytecode.frontend` tests OOM with the default heap. Reproduced on clean `develop`, so not from this PR, but it may be worth a separate look.
- `.parallel()` in `getClassSource` is pointless for a short input-location list, but measured as noise (1.03×) — left alone.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sootup.java.core/src/main/java/sootup/java/core/views/EagerLoadingStrategy.java
+++ b/sootup.java.core/src/main/java/sootup/java/core/views/EagerLoadingStrategy.java
@@ -1,0 +1,79 @@
+package sootup.java.core.views;
+
+/*-
+ * #%L
+ * SootUp
+ * %%
+ * Copyright (C) 2026 Markus Schmidt and others
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, either version 2.1 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Lesser Public License for more details.
+ *
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-2.1.html>.
+ * #L%
+ */
+
+import java.util.Optional;
+import org.jspecify.annotations.NonNull;
+import sootup.core.inputlocation.AnalysisInputLocation;
+import sootup.core.types.ClassType;
+import sootup.java.core.JavaSootClass;
+
+/**
+ * Resolves every class of a {@link JavaView} up front, at construction time, then releases the
+ * view's input locations. Used by {@link JavaEagerView}.
+ *
+ * <p>Since {@link #initialize} has already resolved (and cached) everything reachable, a class not
+ * found in the cache by the time {@link #resolveOnCacheMiss} runs will never be found - there is
+ * nothing left to probe, and the input locations are closed besides. This carries no absence-set
+ * bookkeeping, unlike {@link OnDemandLoadingStrategy}, because none is needed.
+ */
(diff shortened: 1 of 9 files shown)


swissiety (commented):
Good idea. How about extracting this into a configurable Strategy pattern for Loading in  the View: 
1) ON_DEMAND
2) EAGER

Then the bookkeeping is only necessary in ON_DEMAND as EAGER would have already resolved everything in advance.

The current implementation for eager loading contains an EagerJavaView which loads all available classes in the constructor. Unfortunately its done via inheritance at the moment - composition e.g. a Strategy pattern would be preferred.

Possible downside compared to the current PR changes: No default tracking of absent classes in ON_DEMAND but this could be easily extended if someone is interested in that information.

swissiety (approved):
👍

PR #1727, opened 2026-09-02
Fix: NullPointerException in exception analysis for new-array expressions with a primitive base type

solid-eureka:

`StmtExceptionAnalyzer::mightThrow()` throws a `NullPointerException` for any array creation with a primitive base type, e.g. `int[] arr = new int[4];`.

`ExceptionInferExprVisitor.caseNewArrayExpr()` only initialized its `result` field when the array's base type is a `ReferenceType`:

```java
public void caseNewArrayExpr(@NonNull JNewArrayExpr expr) {
  if (expr.getBaseType() instanceof ReferenceType) {
    result = new ExceptionInferResult(ErrorType.RESOLVE_CLASS_ERROR);   // <-- only here
  }
  Value count = expr.getSize();
  if (count instanceof Local) {
    result = result.addException(NEGATIVE_ARRAY_SIZE_EXCEPTION, hierarchy);
  } else if (count instanceof IntConstant) { ... }
}
```

Unlike `ExceptionInferStmtVisitor`, this visitor's constructor never seeds `result`, and every other `case*` method assigns it unconditionally. So for a primitive base type `result` stays `null`, which gives two failure modes:

* **Constant size** (`new int[4]`) — `getResult()` returns `null` and the NPE surfaces one frame later in `ExceptionInferResult.addExceptions` on `newResult.exceptions`:

  ```
  java.lang.NullPointerException: Cannot read field "exceptions" because "newResult" is null
      at sootup.java.core.exceptions.ExceptionInferResult.addExceptions(ExceptionInferResult.java:94)
      at sootup.java.core.exceptions.ExceptionInferStmtVisitor.caseAssignStmt(ExceptionInferStmtVisitor.java:85)
      at sootup.java.core.exceptions.StmtExceptionAnalyzer.mightThrowImplicitly(StmtExceptionAnalyzer.java:112)
  ```

* **Variable size** (`new int[a]`) — NPE directly inside `caseNewArrayExpr` on `result.addException`.

The existing `StmtExceptionAnalyzerTest` did not catch this because its test resource only creates reference-typed arrays (`new Integer[5]`).

`result` is now assigned unconditionally, so the branch below can never dereference `null`:

```java
result =
    expr.getBaseType() instanceof ReferenceType
        ? new ExceptionInferResult(ExceptionInferResult.ErrorType.RESOLVE_CLASS_ERROR)
        : ExceptionInferResult.createEmptyException();
```

The `createEmptyException()` for the primitive case should be correct, because there can't be a RESOLVE_CLASS_ERROR for a primitive type.

`caseNewMultiArrayExpr` needs no equivalent change, because the base type of `JNewMultiArrayExpr` is always an `ArrayType` and thus a `ReferenceType`.

Two test methods were added to `StmtExceptionAnalyzerTest`, built directly from the Jimple API so no new binary test resource is needed:

| statement | expected exceptions |
| --- | --- |
| `new int[4]` | `VirtualMachineError`, `ThreadDeath` |
| `new int[-1]` | + `NegativeArraySizeException` |
| `new int[size]` | + `NegativeArraySizeException` |
| `new Integer[4]` | + `LinkageError` (regression guard) |

Both fail with the NPE above before the fix and pass after it. The three pre-existing tests in the class are unaffected.

**Linked issue (if any)**

None. I did not find an existing issue for this.

**Checklist**

*Code style & guidelines*
- [ ] I updated documentation (if needed) — not needed, no public API or documented behaviour changes

*Self-review*
- [x] I have successfully run tests with your changes locally (`sootup.tests` `StmtExceptionAnalyzerTest`: 5/5 pass; full `sootup.java.core` suite passes)

*Review*
- [ ] CI checks are green

### DIFF
--- a/sootup.java.core/src/main/java/sootup/java/core/exceptions/ExceptionInferExprVisitor.java
+++ b/sootup.java.core/src/main/java/sootup/java/core/exceptions/ExceptionInferExprVisitor.java
@@ -223,9 +223,11 @@ public class ExceptionInferExprVisitor extends AbstractExprVisitor {
 
   @Override
   public void caseNewArrayExpr(@NonNull JNewArrayExpr expr) {
-    if (expr.getBaseType() instanceof ReferenceType) {
-      result = new ExceptionInferResult(ExceptionInferResult.ErrorType.RESOLVE_CLASS_ERROR);
-    }
+    // only a reference base type has to be resolved, a primitive one cannot cause a LinkageError
+    result =
+        expr.getBaseType() instanceof ReferenceType
+            ? new ExceptionInferResult(ExceptionInferResult.ErrorType.RESOLVE_CLASS_ERROR)
+            : ExceptionInferResult.createEmptyException();
     Value count = expr.getSize();
     if (count instanceof Local) {
       result =
--- a/sootup.tests/src/test/java/sootup/tests/exceptions/StmtExceptionAnalyzerTest.java
+++ b/sootup.tests/src/test/java/sootup/tests/exceptions/StmtExceptionAnalyzerTest.java
@@ -6,7 +6,11 @@ import java.util.*;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import sootup.core.interceptor.BodyInterceptor;
+import sootup.core.jimple.Jimple;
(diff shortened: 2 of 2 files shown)


swissiety (approved):
👍 thx @solid-eureka for your contribution!

PR #1728, opened 2026-09-03
Unreachable block parsing bug in the ApkFrontend

Hazem-Gamall:

`DexBody` converts the dex instruction list linearly, so every instruction becomes a `Stmt` whether or not control can reach it. Dalvik allows plenty that cannot be: the `nop` padding a switch or fill-array-data payload to a 4-byte boundary, and anything an obfuscator parks behind a `goto`. Those Stmts end up in blocks nothing points at, and `Body.build` rejects the graph:

```
Stmt 'nop' which is neither the StartingStmt nor a TrapHandler is missing a predecessor!
```

1601 of 6463 classes in one obfuscated APK failed to load. The JVM frontend never hits this because it converts along the flow and never visits an unreachable instruction.

Two changes, neither sufficient alone:

1. **A `Stmt` that does not fall through ends its block.** `blockEnd` previously held only branching Stmts, so a `return` was chained to whatever followed it and inherited its block's fall-through edge (`return: must have '0' outgoing flow but has '1'`). Dead code now lands in its own block instead of hiding inside a live one.

2. **`removeUnreachableBlocks`**  a worklist from the entry block over fall-through and branch edges. A handler is marked once any block in its trap's range is, which can open new flow, so the two alternate to a fixpoint. Unreachable blocks are dropped along with their `successorMap` entries; traps are dropped when their begin or handler is gone, and their end advanced when only that was removed.

Reachability is the criterion, not the specific opcode.

Measured over 7 APKs: 280,314 bodies convert, 0 failures.

**Linked issue (if any)**

**Checklist**

*Code style & guidelines*
- [x] I added the necessary comments

*Self-review*

*Review*
- [ ] CI checks are green

### DIFF
--- a/sootup.apk.frontend/src/main/java/sootup/apk/frontend/main/DexBody.java
+++ b/sootup.apk.frontend/src/main/java/sootup/apk/frontend/main/DexBody.java
@@ -378,10 +378,138 @@ public class DexBody {
     if (!currentList.isEmpty()) {
       listList.add(currentList);
     }
+    removeUnreachableBlocks(listList, branchingStmtListMap);
     graph.initializeWith(listList, branchingStmtListMap, traps);
     DexMethodSource dexMethodSource =
         new DexMethodSource(locals, methodSignature, graph, method, bodyInterceptors, view);
     return dexMethodSource.makeSootMethod();
+  }
+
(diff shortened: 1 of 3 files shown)


swissiety (approved):
👍

PR #1729, opened 2026-09-03
Exception handler merge corruption effecting Dex bodies

Hazem-Gamall:
> Builds on #1728, this branch contains its commits, so review that one first. The changes here are confined to `sootup.core`.

  `insertBefore` on the head of a Trap handler corrupts the graph. A block keeps one flat list of who points at it, and both normal and exceptional edges write to it. `insertBefore` moved *every* entry onto the inserted block, but `replaceSuccessorBlock` only rewrites the normal successor array; so an exceptional edge kept pointing at the old block while the entry recording it was deleted. The handler then looked unreachable and `tryMergeBlocks` merged it away, leaving the catching blocks pointing at a block `getBlocks()` no longer holds.

  `BlockGraphIterator` follows exceptional successors into that stale block and returns more blocks than the graph has:

  ```
  There are -3 Blocks that are not iterated! i.e. the ControlFlowGraph is not connected from its startingStmt!
  ```

  A negative count means a duplicated block, not a missing one.

  Reaching this needs a handler whose first `Stmt` is an ordinary one, a legal `insertBefore` target. That is common in dex and impossible on the JVM:

  - In Dalvik, `move-exception` is optional, it "must be the first instruction of any exception handler ***whose caught exception is not to be ignored***" ([Dalvik bytecode][dalvik], opcode `0d`). Ignore the exception and there is no `move-exception`; the handler simply starts executing.
  - On the JVM the exception is always handed to the handler on the operand stack, "the operand stack of the current frame is cleared, *objectref* is pushed back onto the operand stack, and execution continues" ([JVMS §6.5 `athrow`][athrow]); so the handler is forced to consume it with an `astore` or a `pop`.

  `AsmMethodSource` therefore always has something to model and always emits `@caughtexception`. `DexBody.addTraps` does not, and falls back to the handler instruction's own `Stmt`, logging `First instruction of trap handler unit not MoveException but ...`. Measured: **741 of 2795** traps in one APK and **1995 of 6110** in another have a handler that is not a `@caughtexception` `Stmt`. `BoomerangPreInterceptor` and `TypeAssigner` both insert before such heads.

The failure was also unreadable: building the exception message re-entered the failing iterator, so it surfaced as a `StackOverflowError` instead.

[dalvik]: https://source.android.com/docs/core/runtime/dalvik-bytecode
[athrow]: https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-6.html#jvms-6.5.athrow

1. **`insertBefore`** re-points the exceptional edges onto the inserted block rather than abandoning them, everything that entered the old block must enter the inserted `Stmt`s, or they are unreachable. Predecessor entries are moved by count, since a predecessor can reach a block through several edges at once (two case labels of one switch, or a normal edge beside an exceptional one) and the entries are untagged.

2. **`isMergeable`** refuses to merge away a block an exceptional edge enters. `insertBefore` handles its own case, but `removeNode` and `tryMergeWithSuccessorBlock` reach `tryMergeBlocks` too, and there the caller cannot know whether `firstBlock`'s statements belong on the exceptional path. Merging is only a layout optimisation, so refusing costs one extra block.

Fix 1 alone left 1 and 3 failures on two APKs; both together, 0.

3. **`ReversePostOrderBlockTraversal`** sized its list with `cfg.getStmts()`, which walks the graph through the validating iterator. Sized from `getBlocks().size()` instead, so the real `IllegalStateException` surfaces rather than a `StackOverflowError`.

Two tests in `MutableBlockControlFlowGraphTest` cover a handler reached both exceptionally and normally, and one reached only exceptionally. Both fail on `develop` with `an exceptional edge points at a block the graph does not hold anymore`.

**Linked issue (if any)**

Builds on #1728.

**Checklist**

*Code style & guidelines*

*Self-review*

*Review*
- [ ] CI checks are green

### DIFF
--- a/sootup.apk.frontend/src/main/java/sootup/apk/frontend/main/DexBody.java
+++ b/sootup.apk.frontend/src/main/java/sootup/apk/frontend/main/DexBody.java
@@ -378,10 +378,138 @@ public class DexBody {
     if (!currentList.isEmpty()) {
       listList.add(currentList);
     }
+    removeUnreachableBlocks(listList, branchingStmtListMap);
     graph.initializeWith(listList, branchingStmtListMap, traps);
     DexMethodSource dexMethodSource =
         new DexMethodSource(locals, methodSignature, graph, method, bodyInterceptors, view);
     return dexMethodSource.makeSootMethod();
+  }
+
+  private void removeUnreachableBlocks(
+      List<List<Stmt>> blocks, Map<BranchingStmt, List<Stmt>> successorMap) {
+    if (blocks.isEmpty()) {
+      return;
+    }
+
+    Map<Stmt, Integer> blockOfHead = new IdentityHashMap<>();
+    for (int i = 0; i < blocks.size(); i++) {
+      blockOfHead.put(blocks.get(i).get(0), i);
+    }
+
+    boolean[] reachable = new boolean[blocks.size()];
+    Deque<Integer> worklist = new ArrayDeque<>();
+    reachable[0] = true;
+    worklist.add(0);
+    // A handler is entered by an exceptional edge from the Stmts its Trap covers, so it becomes
+    // reachable as soon as any block of that range is.
+    boolean foundHandler = true;
+    while (foundHandler) {
+      followEdges(blocks, successorMap, blockOfHead, reachable, worklist);
+      foundHandler = false;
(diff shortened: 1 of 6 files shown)


swissiety (approved):
👍 good job

PR #1745, opened 2026-09-21
enable identifier hash-consing

swissiety:
hide ctor of `ClassType`,`MethodSignature`  etc. to enfore the usage of the IdentifierFactory. That change allows us to use == and IdentityHashMaps for these objects in the future to improve performance.

**Checklist**
[x] yes


sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
```diff
@@ -516,11 +516,12 @@ protected void implicitStartRunCall(
...
-              methodSig.getDeclClassType(),
-              "run",
-              methodSig.getParameterTypes(),
-              methodSig.getType());
+          view.getIdentifierFactory()
+              .getMethodSignature(
+                  methodSig.getDeclClassType(),
+                  "run",
+                  methodSig.getType(),
+                  methodSig.getParameterTypes());
```
JonasKlauke, line 524: Is this correct?
isnt it params and then return type?

JonasKlauke (approved):

sootup.callgraph/src/main/java/sootup/callgraph/AbstractCallGraphAlgorithm.java
swissiety: nope

PR #1753, opened 2026-09-21
* fix(core): exclude statement itself when searching reaching definitions in Local.getDefsForLocalUse

dkimitsa:
Local.getDefsForLocalUse(graph, stmt) determines the reaching definition statements for a local variable used at a given statement stmt. The BFS traversal queue was previously initialized with stmt itself (queue.add(stmt)). If stmt is an assignment that also defines the same local variable (e.g. x = x + 1), the search encountered stmt.getDef() on its very first pop, added stmt to the reaching definitions, and executed continue, skipping traversal of stmt predecessors.

This creates cyclical, self-referential def-use relationships where an expression claims to be defined by its own assignment before that assignment has taken place. This breaks constant propagation, dead code elimination, and alias analysis passes that depend on accurate reaching definitions, causing them to treat variables as uninitialized or incorrectly initialized.

1. Initialized the BFS queue with graph.predecessors(stmt) instead of stmt.
2. This searches all inbound control flow edges reaching the statement without falsely treating the statement own definition as reaching its own operands.
3. If the statement is part of a loop (e.g. back-edge from a later iteration), the statement definition can still be reached via traversal of the predecessor back-edge.

**Steps to Reproduce**

1. Source for Subject File
```java

public class SelfDefinition {
    public static int increment() {
        int x = 5;
        x = x + 1;
        return x;
    }
}
```

2. SootUp Test Code to Reproduce
```java
    JavaView view = new JavaView(inputLocation);
    JavaIdentifierFactory idFactory = view.getIdentifierFactory();
    JavaClassType classType = idFactory.getClassType("example.SelfDefinition");
    JavaSootClass sootClass = view.getClass(classType).orElseThrow();
    JavaSootMethod method = sootClass.getMethod("increment", Collections.emptyList()).orElseThrow();

    Body body = method.getBody();
    Body.BodyBuilder builder = Body.builder(body, Collections.emptySet());
    ControlFlowGraph<?> cfg = builder.getControlFlowGraph();
    Stmt s0 = body.getStmts().get(0);
    Stmt s1 = body.getStmts().get(1);
    Local x = (Local) s0.getDef().get();
    System.out.println("Testing getDefsForLocalUse on stmt: " + s1);
    System.out.println("Defs for local x: " + x.getDefsForLocalUse(cfg, s1));
    // before the fix: Defs for local x: [x = x + 1]
    // after the fix: Defs for local x: [x = 5]
```

**Checklist**

*Code style & guidelines*

*Self-review*
- [x] I have successfully run tests with my changes locally


swissiety (approved):
great finding @dkimitsa 👍

PR #1761, opened 2026-09-21
* fix (core): allow empty trap range in Trap constructor and withers

dkimitsa:

Trap's constructor throws IllegalArgumentException ("The covered Trap range is empty. Trap is of no use.") whenever beginStmt == endStmt. 

Why It Broke Transformations:
1. Compiler Passes: Dead-code elimination, inlining, and statement pruning can collapse all protected statements or redirect endpoints to the same successor statement.
2. Wither Methods: .withBeginStmt (...) and .withEndStmt (...) delegate to new Trap (...). Any intermediate or permanent step setting beginStmt == endStmt caused transformation pipelines to crash.

Removed the beginStmt == endStmt exception check, allowing empty trap ranges to exist and be manipulated safely.

**Checklist**

*Code style & guidelines*

*Self-review*
- [x] I have successfully run tests with my changes locally


swissiety:
This behaviour is intentional. E.g. If a transformer removed all stmts in the trap range it needs to cleanup the leftover catch clause as well instead of doing only half the work. Feel free to reopen this PR if you want to discuss further.

